PR #42, opened 2021-10-26
[perf][refactor] Fused linear rewrite

### LINKED ISSUE #73: [bug] Flaky FusedMLP test
# 🐛 Bug
See https://app.circleci.com/pipelines/github/facebookresearch/xformers/303/workflows/5b19d79c-572d-46c5-a3af-66e65e04385d/jobs/575
Could be a good incentive to close on https://github.com/facebookresearch/xformers/pull/42

blefaudeux:
## What does this PR do?
Focus on improving the speed for this MLP option, and clean up the logic. Should incidentally fix #73

The logic now is that:
- FW is fused, we save the activation inputs if required
- BW fires 4 kernels, 3 of which being standard matmul or reduction, the last one being a simple coefficient wise kernel which happens to fuse the backpropagation

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


blefaudeux:
the plots are old at the moment, regenerating new ones

blefaudeux:
I cannot repro the segfaults locally, I'm not sure whether it could be that it's a T4+triton issue

xformers/triton/k_fused_matmul_bw.py
```diff
@@ -0,0 +1,149 @@
...
+import triton.language as tl
+
+
+# fmt: off
+@triton.heuristics({
+    'EVEN_N': lambda *args, **meta: args[3] % (meta['BLOCK_COL']) == 0,
+})
+@triton.autotune(
+    configs=[
+        triton.Config({"BLOCK_COL": 32}, num_stages=5, num_warps=2),
```
blefaudeux, line 20: @zdevito, this will be an input to the MLP for instance

xformers/triton/k_fused_matmul_fw.py
```diff
@@ -0,0 +1,194 @@
...
+import triton.language as tl
+
+# CREDITS: Initially inspired by the Triton tutorial on matrix multiplications
+
+
+# fmt: off
+@triton.heuristics({
+    'EVEN_K': lambda *args, **meta: args[7] % (meta['BLOCK_K']) == 0,
+})
+@triton.autotune(
```
dianaml0, line 19: Just curious, are these values from testing and seeing which configs help?

xformers/triton/activations.py
```diff
@@ -68,7 +68,12 @@ def relu(x):
 
 @triton.jit
 def relu_grad(x):
-    return tl.where(x >= 0, 1.0, 0.0)
+    # ReLU is different from other activations
+    # in that it does not require the input to retrospectively compute its gradient
+    # here the input is the downstream gradient, and we return the upstream gradient directly
+    zero = 0.0
+    zero = zero.to(x.dtype)
```
dianaml0, line 75: Should this be done for zero in `squared_relu_grad` as well?

xformers/triton/k_fused_matmul_bw.py
```diff
@@ -0,0 +1,149 @@
...
+        triton.Config({"BLOCK_COL": 64}, num_stages=5, num_warps=2),
+        triton.Config({"BLOCK_COL": 128}, num_stages=3, num_warps=4),
+        triton.Config({"BLOCK_COL": 256}, num_stages=3, num_warps=8),
+        triton.Config({"BLOCK_COL": 512}, num_stages=3, num_warps=8),
+        triton.Config({"BLOCK_COL": 1024}, num_stages=3, num_warps=16),
+    ],
+    key=["N"],
+)
+@triton.jit
+def kernel_bw(
```
dianaml0, line 30: Really helpful to have a short description at every step!

dianaml0 (approved):
🚀

xformers/triton/activations.py
blefaudeux: it's a bit confusing, but it works without that because squared_relu ends up getting the saved activation input, which is fp32 by default (whereas with AMP the downstream gradient will be fp16). It makes me think that maybe there's some place to save here by casting it to fp16 when the inputs are fp16, will also improve the perf a little, that's a good point !

xformers/triton/k_fused_matmul_fw.py
blefaudeux: ah I need to explain that, it's really hard to grasp.. args[7] relates to the kernel inputs below, it's just the dimension "K" (the dimension against which the matrix mul is consolidated). This "heuristic" decorator states that if K is divisible by BLOCK_K (triton which test out the presets above for this value, best will depend on HW and inputs), then there's no need to mask the loads in the kernels (this `tl.load(.., maskl=())`), we will not overreach anyway.

blefaudeux:
Perf is not the best since I finally understood the reason for the crashes on T4, the triton kernel was using too much shared memory (which is normally caught by outofressourceserror, but not this time). Falling back to smaller configs for now, could be improved subsequently by having per-HW configs, which we had at some point (not too hard to do)

PR #53, opened 2021-10-29
[feat] InplaceNorm, affine-less LayerNorm

### LINKED ISSUE #50: feat(triton): InplaceNorm + InstanceNorm
I'd love to run LayerNorm in place and ideally also add InstanceNorm (by extracting the core normalization from LayerNorm) as HomebrewNLP is currently using a slow PyTorch-level [implementation](https://github.com/HomebrewNLP/HomebrewNLP/blob/master/src/model.py#L32-L53) with a correct backward pass.

While we're at it, optionally fusing GLU and [GLUv2](https://arxiv.org/abs/2102.11856) (`gelu(f(x)) * g(x) + gelu(h(x))`) with various activation functions and normalization might give another speed boost.

To add this myself, I'd need to fully understand triton's pointers and how to access the output instead of input in your LayerNorm implementation. Could you help me with that? or would you instead implement this yourself? Is this even in the scope of xformers?

ClashLuke:
This should resolve #50 

- [x] InplaceNorm
- [x] affine-less LayerNorm
- [ ] GLU


blefaudeux:
Else the results are pretty great @ClashLuke, and composing the transforms while the data is in shared memory is the right thing to do I think

xformers/triton/k_layer_norm.py
```diff
@@ -246,38 +323,44 @@ def backward(ctx, dy):
...
         # we suppose in the kernel that they have the same size and are contiguous
-        assert dx.numel() == x.numel(), \
+        assert dx.numel() == y.numel(), \
             "Something is wrong in the backward graph, possibly because of an inplace operation after the layernorm"
 
         # enqueue kernel using forward pass heuristics
         # also compute partial sums for DW and DB
 
         # fmt: off
+        meta = {"BLOCK_SIZE_N": ctx.BLOCK_SIZE_N,
```
blefaudeux, line 348: you probably saw that, but this is the same as passing it directly down when calling the kernel

blefaudeux (approved):
Provided you rebase out of the other PR, this looks great to me, thanks a lot @ClashLuke. I'd like to homogenize the coding style on all the kernels so that this part is easier to grasp, but I'll do that after this lands

blefaudeux:
@ClashLuke ok to 
- rebase out of the reversible branch ? (you can 'git rebase --onto main {hash of the commit just before the stack you want to move}')
- implement the GELU path in another PR ? could be that I can homogenize the kernels before that

xformers/triton/k_layer_norm.py
ClashLuke: I could integrate it into the two calls below. It just felt cleaner to not repeat shared arguments.

ClashLuke:
It doesn't look like rebasing solved the problem as I merged inside of the graph, so I just squashed the changes into a single commit.

blefaudeux:
@ClashLuke  alright, looks good to me, you would just need to rebase on main now :) (sorry about all that, the doc build problem was fixed yesterday)

blefaudeux:
Looks good, I'll land as soon as the final test is done, thank you so much @ClashLuke ! Adding you to the contributors in a future README update

Issue #43, opened 2021-10-27
Support shape types for Tensors.


pradeep90:
# 🚀 Feature

Add shape type stubs for Tensor functions. Annotate some of the xformers code with shape types so that future users can add new ones easily. Set up IDE support and iterate on the UX based on user feedback.

## Motivation

Shape types allow us to see the shape of a Tensor as it is transformed by various functions. This can help understand existing code, debug problems, and catch shape mismatches without running the entire program. 

For example, it can catch broadcasting errors like `torch.randn(2, 3) + torch.randn(4, 5)`.

This requires annotating Tensor variables with their initial shapes. More docs to come.

## Pitch

We should be able to see the shape of a Tensor variable as we code (after adding annotations to the function signature).

## Alternatives

The existing alternatives are to `print(my_tensor.shape)`, try it out in a notebook, or run the program end-to-end.

## Additional context

PEP 646: Variadic types: https://www.python.org/dev/peps/pep-0646/

More docs to come.

PR #63, opened 2021-11-01
[feat] Add shape types to sine.py.

pradeep90:
The main changes are:

+ Make the class generic in a variable `DimModel` and set the type of
`self.dim_model` as `DimModel`. That way, when it is initialized with
`self.dim_model`, Pyre will store the type. We have to add
`Generic[DimModel]` in this class because `self.dim_model` is defined in
this class.

+ Annotate `forward` as accepting a Tensor of shape N x N and returning
a Tensor of shape N x N x DimModel.

This is about as straightforward as it gets for shape types. There were no
major complications in this class. If even this is too cumbersome or hard to
follow, it will not get easier in other classes :) So, feedback welcome on the
understandability.

Note:

+ I added `from __future__ import annotations` so that the type
annotations are not evaluated at runtime. This is necessary because the
runtime Tensor class is not generic and cannot take arguments like
`Tensor[torch.float32, N]`, etc.

+ I changed `torch.arange(0.0, seq_len)` to `torch.arange(0, seq_len,
dtype=torch.float32)` because Pyre won't be able to calculate the
resulting Tensor shape using floats like 0.0. It is able to do so with
integers like 0. Note that I can't make it `dtype=x.dtype` because 
`x.dtype` can sometimes be int, as in the unit tests.

+ An inconvenience is that we check if x.ndim == 2 and then unsqueeze
the last dimension. However, this means that there could be two possible
shapes for x, whereas we have declared that it has exactly shape N x N.
So, I added a local variable `output` with the desired shape after
unsqueezing and suppressed the Pyre error with a comment.

I'll try to see if Pyre can allow such idioms in the future.

## What does this PR do?
Works towards #43 

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [x] N/A
  - [x] N/A
  - [x] N/A

## PR review

### DIFF
--- a/stubs/torch_stub_tests.py
+++ b/stubs/torch_stub_tests.py
@@ -700,11 +700,8 @@ def test_argsort() -> None:
 def test_functional_pad() -> None:
     x: torch.Tensor[torch.float32, L[2], L[3], L[4]]
 
-    # pyre-ignore[9]: Fixed in the latest version of Pyre.
     good: torch.Tensor[torch.float32, L[2], L[3], L[5]] = nn.functional.pad(x, (1, 0))
-    # pyre-fixme[9]: bad has type `Tensor[torch.float32, typing_extensions.Literal[99...
     bad: torch.Tensor[torch.float32, L[99], L[3], L[5]] = nn.functional.pad(x, (1, 0))
-    # pyre-ignore[9]: Fixed in the latest version of Pyre.
     good2: torch.Tensor[torch.float32, L[2], L[10], L[7]] = nn.functional.pad(
         x, (1, 2, 3, 4), "constant", value=0.0
     )
@@ -1187,7 +1184,6 @@ def test_repeat_interleave() -> None:
 
     # Too dynamic because the output shape depends on the contents of repeats.
 
-    # pyre-ignore[9]: Fixed in the latest version of Pyre.
     y5: torch.Tensor[torch.float32, L[0], L[3], L[4]] = torch.repeat_interleave(
         x, repeats, dim=0
     )
--- a/xformers/components/positional_embedding/sine.py
+++ b/xformers/components/positional_embedding/sine.py
@@ -4,9 +4,17 @@
 # LICENSE file in the root directory of this source tree.
 
 
+# Silence Mypy errors in this file.
+# type: ignore
+
+from __future__ import annotations
+
 import math
+from typing import TypeVar
 
 import torch
+from pyre_extensions import Generic
+from typing_extensions import Literal as L
 
 from xformers.components.positional_embedding import (
     PositionEmbedding,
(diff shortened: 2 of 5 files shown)


xformers/components/positional_embedding/sine.py
```diff
@@ -4,32 +4,45 @@
 # LICENSE file in the root directory of this source tree.
 
 
+# Silence Mypy errors in this file.
```
pradeep90, line 7: We are still using Mypy, which means it complains about sine.py since it doesn't understand variadics yet.

So, I've made Mypy ignore the entire file. Long-term, if we move off Mypy, we can remove this.

pradeep90:
cc @blefaudeux Could you take a look? I don't seem to be able to add reviewers

xformers/components/positional_embedding/sine.py
```diff
@@ -4,32 +4,45 @@
...
+        self.dim_model: DimModel = dim_model
 
-    def forward(self, x: torch.Tensor):
+    def forward(
+        self, x: torch.Tensor[torch.float, N, N]
+    ) -> torch.Tensor[torch.float, N, N, DimModel]:
         seq_len = x.shape[1]
         pos = (
-            torch.arange(0.0, seq_len, device=x.device)
+            torch.arange(0, seq_len, device=x.device, dtype=torch.float32)
```
blefaudeux, line 40: `torch.arange(0, seq_len, device=x.device, dtype=x.dtype)`, else this will error out if x is fp16

blefaudeux: oh, thinking about it, it's even strange that it works right now

pradeep90: That's what I tried at first, but even the unit test fails because `x` can be `int`. Alternatives welcome. (Or is the unit test misleading?)

blefaudeux: oh of course, sorry, it has to be a float type (inputs are typically tokens, which are int), fp32 should be fine actually (the layers consuming this can always cast it down to fp16). All good

stubs/torch_stub_tests.py
```diff
@@ -700,11 +700,8 @@ def test_argsort() -> None:
 def test_functional_pad() -> None:
     x: torch.Tensor[torch.float32, L[2], L[3], L[4]]
 
-    # pyre-ignore[9]: Fixed in the latest version of Pyre.
```
blefaudeux, line 703: nice for all these

blefaudeux (approved):
LGTM, a little concerned about the mypy / pyre cohabitation though, but looks like an ok tradeoff so far. I need to checkout pyre myself actually, any recommended instructions ?

blefaudeux:
ah I was too fast, this PR basically makes pyre mandatory, since the import would fail. Thoughts @dianaml0 @jieru-hu @fmassa ?

pradeep90:
> ah I was too fast, this PR basically makes pyre mandatory, since the import would fail. Thoughts @dianaml0 @jieru-hu @fmassa ?

Oh. I added `pyre-extensions` to the `requirements-test.txt` file. Do we need to do something else or is the concern that we're adding a dependency?

blefaudeux:
> > ah I was too fast, this PR basically makes pyre mandatory, since the import would fail. Thoughts @dianaml0 @jieru-hu @fmassa ?
> 
> Oh. I added `pyre-extensions` to the `requirements-test.txt` file. Do we need to do something else or is the concern that we're adding a dependency?

there's a basic requirement set in `requirements.txt`, right now it's only pytorch, my concern is that we're force feeding pyre to everyone with that (else the import will fail), but on the other hand I realize that there's not much you can do without that. It's fine by me basically, but I wanted to raise this to the other stakeholders

PR #84, opened 2021-11-07
Automate binary releases

fmassa:
This PR sets up CircleCI jobs that automatically build wheels for xformers, for different Python versions.

In order to kick the builds, the following is needed:
- create a branch named `release/v0.0.2`  (one can change the release version)
- apply needed modifications for the release (if needed)
- create a pull request (which don't need to be merged)

The pull request will launch the build jobs, together with a set of jobs to validate that the extensions have been properly built and can be loaded.

~In order to test this PR, I've temporarily changed the name of the branch to be `binaries_job` (the name of this branch), but last commits in this PR changes it to the corresponding regex for `release/...`~

EDIT: in order to facilitate testing / reviewing, I propose to for now merge this PR with the `binaries_job` in the requirement for branch name, and then send a follow-up PR on top of a branch named `release/0.0.0` with the changes in the regex.

I'll be using the following regex `/release\/[0-9]+(\.[0-9]+)+/` in a follow-up PR

### DIFF
--- a/packaging/build_wheel.sh
+++ b/packaging/build_wheel.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+# Copyright (c) Facebook, Inc. and its affiliates.
+# Taken with modifications from https://github.com/facebookresearch/detectron2/blob/main/dev/packaging/build_wheel.sh
+set -ex
+
+ldconfig  # https://github.com/NVIDIA/nvidia-docker/issues/854
+
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+. "$script_dir/pkg_helpers.bash"
+
+echo "Build Settings:"
+echo "CU_VERSION: $CU_VERSION"                             # e.g. cu101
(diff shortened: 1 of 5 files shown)


packaging/build_wheel.sh
```diff
@@ -0,0 +1,32 @@
...
+echo "Build Settings:"
+echo "CU_VERSION: $CU_VERSION"                             # e.g. cu101
+echo "XFORMERS_VERSION_SUFFIX: $XFORMERS_VERSION_SUFFIX"   # e.g. +cu101 or ""
+echo "PYTHON_VERSION: $PYTHON_VERSION"                     # e.g. 3.6
+echo "PYTORCH_VERSION: $PYTORCH_VERSION"                   # e.g. 1.4
+
+setup_cuda
+setup_wheel_python
+
+yum install ninja-build -y
```
blefaudeux, line 20: assuming redhat/centos/fedora I guess ?

blefaudeux (approved):
Seems pretty great to me ! To be transparent I didn't know half of that, but from what I can see it seems clean and does the job. Thanks a lot @fmassa, this will be super helpful, I think that if pypi is not ok with that we could host the binaries on circleci ('artifacts'), github actions or just the fair endpoints (not familiar with them)

Issue #78, opened 2021-11-05
[feat] patch torch and huggingface for improved performance in existing applications


ClashLuke:
I'd love it if we could patch existing functions like `torch.nn.functional.softmax` with our faster xformers/triton implementation. This would allow users of pre-defined models from HuggingFace to simply call `xformers.patch()` and run twice as fast.\
Do you think that'd be possible, or would we have to rewrite all APIs first to match PyTorch's?

I'd propose writing a `patch()` function which replaces the appropriate PyTorch modules with our implementation at run-time. Alternative implementations are more than welcome, as my proposal wouldn't allow jumping into the source code of the new PyTorch function, rendering this change invisible to the user.

blefaudeux:
There's an example here on how to patch existing models, in this case attention but could be generalized to something else https://github.com/facebookresearch/xformers/blob/main/docs/source/vision_transformers.ipynb. Else I've not checked the pytorch interface for softmax, but it's probably very reasonable and we could well have the same here, very doable

PR #89, opened 2021-11-09
WIP: [feat] patch torch for improved performance in existing applications 

ClashLuke:
Implements #78\ 

- [ ] Run `pre-commit` before merging (WSL + Windows PyCharm is a bad combination here)
- [ ] Test if the code works
- [ ] If possible, add support for HuggingFace patching. (They reinvent the wheel a lot, so it's possible that they don't call torch.nn.Linear etc.).
- [ ] Benchmark performance of HuggingFace and PyTorch models before and after patching (maybe `benchmark_patched.py`?)

Feedback is appreciated, especially for the softmax wrapper.\
Should we add run-time assertions to ensure it's only called with [Batch, Sequence, Sequence] features?

### DIFF
--- a/xformers/triton/patch.py
+++ b/xformers/triton/patch.py
@@ -1,0 +1,41 @@
+import typing
+
+import torch
+import torch.nn
+import torch.nn.functional
+
+from dropout import dropout
+from fused_linear_layer import FusedLinear
+from layer_norm import FusedLayerNorm
+from softmax import softmax, log_softmax
+
+
+def _softmax(input: torch.Tensor, dim: int = -1, dtype=torch.float32):
+    if dim == -1:
+        return softmax(input)
+    return softmax(input.transpose(-1, dim)).transpose(-1, dim).to(dtype=dtype)
+
+
+def _log_softmax(input: torch.Tensor, dim: int = -1, dtype=torch.float32):
+    if dim == -1:
+        return log_softmax(input)
+    return log_softmax(input.transpose(-1, dim)).transpose(-1, dim).to(dtype=dtype)
+
+
+def _dropout(input: torch.Tensor, p: float = 0.5, training: bool = True, inplace: bool = False):
+    if training:
+        return dropout(input, p)
+    return input * p
+
+
+def _layer_norm_wrapper(normalized_shape: typing.List[int], eps: float = 1e-5, elementwise_affine: bool = True,
+                        device=None, dtype=None):
+    return FusedLayerNorm(normalized_shape, elementwise_affine, eps).to(device=device, dtype=dtype)
+
+
+def patch_torch():
+    torch.nn.functional.softmax = _softmax
+    torch.nn.functional.log_softmax = _log_softmax
+    torch.nn.functional.dropout = dropout
+    torch.nn.Linear = FusedLinear
+    torch.nn.LayerNorm = FusedLayerNorm
--- a/xformers/triton/patch.py
+++ b/xformers/triton/patch.py
@@ -30,9 +30,16 @@ def _dropout(
     return input * p
 
 
-def _layer_norm_wrapper(normalized_shape: typing.List[int], eps: float = 1e-5, elementwise_affine: bool = True,
-                        device=None, dtype=None):
-    return FusedLayerNorm(normalized_shape, elementwise_affine, eps).to(device=device, dtype=dtype)
+def _layer_norm_wrapper(
+    normalized_shape: typing.List[int],
+    eps: float = 1e-5,
+    elementwise_affine: bool = True,
+    device=None,
+    dtype=None,
+):
+    return FusedLayerNorm(normalized_shape, elementwise_affine, eps).to(
+        device=device, dtype=dtype
+    )
 
 
 def patch_torch():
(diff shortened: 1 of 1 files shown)


xformers/triton/patch.py
```diff
@@ -0,0 +1,41 @@
+import typing
+
+import torch
+import torch.nn
+import torch.nn.functional
+
+from dropout import dropout
```
erip, line 7: Should these be absolute, e.g., `from xformers.triton.dropout import dropout`?

xformers/triton/patch.py
```diff
@@ -0,0 +1,41 @@
...
+from fused_linear_layer import FusedLinear
+from layer_norm import FusedLayerNorm
+from softmax import softmax, log_softmax
+
+
+def _softmax(input: torch.Tensor, dim: int = -1, dtype=torch.float32):
+    if dim == -1:
+        return softmax(input)
```
dianaml0, line 15: Should this be limited to triton? For example there is sparse softmax as well available [here](https://github.com/facebookresearch/xformers/blob/861493c509de8bfc68895c9b5c58fbbc901d5176/xformers/components/attention/core.py#L96)

ClashLuke: I'm not sure I understand. Do you want to call the `_softmax` function instead? I wrote this wrapper to add transpose(), which we don't need with PyTorch softmax.

xformers/triton/patch.py
ClashLuke: Good catch! I'll change that immediately. Thank you!

erip:
It looks like you need to reformat `patch.py`. Running `black xformers/triton/patch.py` should autoformat.

ClashLuke:
Unfortunately, I can't run `pre-commit` locally. If that's all that's left, could you run it on your machine?

xformers/triton/patch.py
dianaml0: Seems like you're replacing `torch.nn.functional.softmax` with triton softmax only. Just suggesting you may want to also patch with sparse softmax if using sparse tensors or have a fallback when triton is not available, unless the focus of this is just triton? Then the way it is now seems fine.

erip:
You don't need to add a precommit hook. Just install black, run the command, git add, commit, and push it.

xformers/triton/patch.py
```diff
@@ -0,0 +1,41 @@
...
+    return softmax(input.transpose(-1, dim)).transpose(-1, dim).to(dtype=dtype)
+
+
+def _log_softmax(input: torch.Tensor, dim: int = -1, dtype=torch.float32):
+    if dim == -1:
+        return log_softmax(input)
+    return log_softmax(input.transpose(-1, dim)).transpose(-1, dim).to(dtype=dtype)
+
+
+def _dropout(input: torch.Tensor, p: float = 0.5, training: bool = True, inplace: bool = False):
```
erip, line 25:
```suggestion
def _dropout(
    input: torch.Tensor, p: float = 0.5, training: bool = True, inplace: bool = False
):
```

xformers/triton/patch.py
```diff
@@ -0,0 +1,41 @@
...
+    if training:
+        return dropout(input, p)
+    return input * p
+
+
+def _layer_norm_wrapper(normalized_shape: typing.List[int], eps: float = 1e-5, elementwise_affine: bool = True,
+                        device=None, dtype=None):
+    return FusedLayerNorm(normalized_shape, elementwise_affine, eps).to(device=device, dtype=dtype)
```
erip, line 33:
```suggestion
def _layer_norm_wrapper(
    normalized_shape: typing.List[int],
    eps: float = 1e-5,
    elementwise_affine: bool = True,
    device=None,
    dtype=None,
):
    return FusedLayerNorm(normalized_shape, elementwise_affine, eps).to(
        device=device, dtype=dtype
    )
```

erip (changes requested):
change formatting to satisfy black.

xformers/triton/patch.py
ClashLuke: Thank you very much!

erip:
The failure looks unrelated, but I'll take a look. If a maintainer can rerun the tests in the meantime, that would be lovely.

blefaudeux:
> The failure looks unrelated, but I'll take a look. If a maintainer can rerun the tests in the meantime, that would be lovely.

sorry just seeing that now, I somehow missed it (ok, I was in a cabin without electricity at the time, could be related). Let me have a look

blefaudeux:
Hey @ClashLuke thanks for the PR and sorry for having been so long to review, I somehow thought that it was a WIP and forgot to peek into it. I think it's a great idea but it's a bit hard to bullet proof, in that pytorch works absolutely all the time and it's a high bar to clear (the custom parts in xformers all assume the same conventions, which are maintained throughout the repo. Softmax is hardcoded to only ever apply to the last dimension for instance). Provided it's clear what the conventions are and if it's useful to people, I think that it's still great to have this. 

I would move that out of /triton and in the /utils if that makes sense, the /triton subfolder is about the triton implementations and it has some consequences for the code coverage for instance, would that be ok ? Thanks again !

p.s: pre-commit can be installed once ([see](https://pre-commit.com/)), you need to activate it once in the repo via `pre-commit install`, and after that all your commits will be auto linted and formatted, we need to update the contributing guide to show that

Issue #90, opened 2021-11-09
`scaled_dot_product_attention` shouldn't slice `attn_mask`


fmassa:
In https://github.com/facebookresearch/xformers/blob/cc7c8edd6907abdb09d10f60e456e224109a4aca/xformers/components/attention/core.py#L233-L248
the core function `scale_dot_product_attention` was made to accept `attn_masks` which are not of the correct shape.

While this looks like a convenience function, IMO this shouldn't be handled at such a core level as `scaled_dot_product_attention`.
Indeed, this behavior can lead to silent bugs (specially if the current slicing strategy doesn't work, e.g., for images), imposes serious performance penalties (by casting sparse tensors to dense tensors), and as such should be handled at a higher level.

Happy to discuss alternatives

blefaudeux:
yes for the slowness, that was the purpose of the logging.warning, it should not be done ideally. I think that it's a hard tradeoff, on the vision side this probably never happens unless there's a mistake (you should rescale the pictures and be done with it), but on the NLP side it's fairly common with different sentences length. I think that there's value in having something convenient (or a hard crash but with enough explanations and helpers so that it's not a blocker), but it could also be somewhere else in the stack indeed, that's certainly fine by me (also the hard crash + plenty of explanations option)

PR #95, opened 2021-11-11
expand dimension in non sparse case as well

dianaml0:
## What does this PR do?
Expand dimension in non sparse case in `matmul_with_mask`

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/xformers/components/attention/core.py
+++ b/xformers/components/attention/core.py
@@ -89,6 +89,8 @@ def _matmul_with_mask(
     # Non optimized codepath
     att = a @ b
     if mask.dtype == torch.bool:
+        if mask.ndim == 2:
+            mask = mask.unsqueeze(0).expand(att.shape[0], -1, -1)
         # mask is presumed false == ignore
         att[~mask] = float("-inf")
     else:


xformers/components/attention/core.py
```diff
@@ -89,6 +89,8 @@ def _matmul_with_mask(
     # Non optimized codepath
     att = a @ b
     if mask.dtype == torch.bool:
+        if mask.ndim == 2:
```
blefaudeux, line 92: it does not work automatically given the broadcasting rules ?

dianaml0: Yea I was surprised broadcasting wasn't working as well, but I was getting this error: 
Screen Shot 2021-11-11 at 4 15 24 PM

blefaudeux: ahh, it's because we index explicitly here, so I'm guessing that broadcasting does not apply !

blefaudeux (approved):
LGTM! I think that @fmassa was thinking about moving these steps out of here though, may be a good idea to keep core/ lean and clean, but it applies to the rest of it

dianaml0:
> LGTM! I think that @fmassa was thinking about moving these steps out of here though, may be a good idea to keep core/ lean and clean, but it applies to the rest of it

That seems like a good idea. Some of the expansion happens in the `ScaledDotProductAttention` vs some happening in `_matmul_with_mask` which could be confusing

blefaudeux:
see https://github.com/facebookresearch/xformers/issues/90, we could move all of that out of this function as @fmassa recommends in a dedicated function, and call it only in the relevant places

PR #98, opened 2021-11-12
[bugfix] Favor + fp16

### LINKED ISSUE #96: Favor breaks with fp16
# 🐛 Bug
The loss nans immediately

## Command
`python3 microGPT.py` when precision is set to 16

## Environment
Unrelated

blefaudeux:
## What does this PR do?
Fixes #96
There was no bug per say, the Softmax approximation for Favor just requires a big dynamic range and the attention computation would overflow in fp16. I tested a couple of takes (limit the values a posteriori, catch inf etc..) but this would be ugly and break the backward graph, in the end keeping this nugget of computations in fp32 is the safest bet I think. Note that we only promote fp16 tensors, not bfloat16 for instance

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/examples/microGPT.py
+++ b/examples/microGPT.py
@@ -28,7 +28,7 @@ class GPT(pl.LightningModule):
         vocab_size,
         weight_decay=0.1,
         betas=(0.9, 0.95),
-        learning_rate=6e-4,
+        learning_rate=1e-4,
         n_embd=512,
         block_size=128,
         n_layer=4,
--- a/examples/microGPT.py
+++ b/examples/microGPT.py
@@ -298,7 +298,7 @@ if __name__ == "__main__":
     model = GPT(
         vocab_size=train_dataset.vocab_size,
         block_size=train_dataset.block_size,
-        attention="nystrom",
+        attention="favor",
         warmup_tokens=REF_BATCH * WARMUP,
         final_tokens=EPOCHS * len(train_dataset) * BLOCK,
     )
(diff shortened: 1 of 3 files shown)


blefaudeux:
cc @xwhan, probably too late for you but FYI works fine now

examples/microGPT.py
```diff
@@ -28,7 +28,7 @@ def __init__(
         vocab_size,
         weight_decay=0.1,
         betas=(0.9, 0.95),
-        learning_rate=6e-4,
+        learning_rate=1e-4,
```
blefaudeux, line 31: before this PR favor cannot run in fp16 (the very first FW dies). With this PR it runs just fine, but it requires a LR a little lower than other mechanisms to prevent overflow later on

dianaml0 (approved):
Interesting bug, LGTM!

blefaudeux:
> Interesting bug, LGTM!

yeah it was really perplexing, k_prime and q_prime are completely fine, then compute att and bam, NaNs and inf all over the place... That's where the fp16 dynamic bites, 65k is not a lot before overflowing. I tried to post fix that in the beginning (knowing that it overflew but that the actual result was fine out of principle), but turns out that it's not completely trivial, not very clean, and not differentiable, so fp32 looked like the safest bet

PR #100, opened 2021-11-13
[minor] Dedicated strided sum kernel

blefaudeux:
## What does this PR do?
Improves the FusedDropoutBias speed, turns out that `torch.sum()` can be suboptimal in that case (we're summing on the dimension which is not typically the contiguous one).

Tons of updated numbers, hard to show everything here, but basically fused dropout and fused linear layers are now faster across the board, sometimes much faster (the plots are in the changed files, all the new plots on a laptop 3080 -*previous plots were sometimes on a V100, hence a possible scale discrepancy, focus on the triton/pytorch gap-*). It's still kind of a mystery for me how to speed up the sum for very big dimensions (> 4096 pytorch becomes faster again), checking with @ptillet

Not too important because not life changing, but faster is always better I suppose.

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


xformers/triton/dropout.py
```diff
@@ -104,7 +104,20 @@ def grid(meta):
         # fmt: on
 
         if ctx.trainable_bias:
-            grad_bias: Optional[torch.Tensor] = torch.sum(grad_in, dim=0)
+            grad_bias: Optional[torch.Tensor] = torch.empty(
```
blefaudeux, line 107: @jieru-hu turns out torch.sum() is not the fastest take out there

blefaudeux:
Screen Shot 2021-11-12 at 4 20 37 PM
back to draft, looks like there's a scheduling issue (should be easy to fix). Benchmarks for the win !

blefaudeux:
two examples of changes (mind the gap, left was on a V100 and right is on a laptop ampere, but the difference with pytorch is visible)

Screenshot from 2021-12-03 23-13-57
Screenshot from 2021-12-03 23-13-22

blefaudeux:
Discussing with @Chillee on another channel, looks like this could be significantly sped up. Looks like we could also give NVFuser a go, it shows very good perf

blefaudeux:
reopening for review @dianaml0 @fmassa, quickly testing the suggestions from Horace on another PR which would end up on top of this one. In particular if the scheduling is changed on the BW pass then the grad bias can be fused altogether with the input grad computations, which could be beneficial (this path seems to be very fast with nvfuser)

blefaudeux:
minor bugfix, ready for review @dianaml0 @fmassa, should be relatively side effect-free. It's still not ideal for dropout, in that the BW pass can be completely fused in that case, I've worked on that on another branch which shows even better speed (basically up to bandwidth in the forward pass (20% better than this PR, give or take), and a bit better than this PR for the BW), but not completely ready

blefaudeux:
Guessing that this is fine, small enough and no comments..

dianaml0:
Sorry I couldn't finish reviewing this earlier @blefaudeux but interesting PR to go through! The comments really helped :)

PR #104, opened 2021-11-15
[bugfix] Causal mask being skipped if no extra mask

### LINKED ISSUE #103: [bug] Causal mask is skipped if no att_mask is passed
# 🐛 Bug

In https://github.com/facebookresearch/xformers/blob/main/xformers/components/attention/scaled_dot_product.py#L78
if no att_mask is passed, then an hypothetical causal mask is completely skipped

blefaudeux:
## What does this PR do?
- Fixes #103
- Adding a unit test checking that causality is correct for scaled dot product and FAVOR. Nystrom does not seem to pass that
- Fixes Favor not picking up the causal flag
- Fixes Favor causal not being trainable (but introduces a big memory use)

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/tests/test_attentions.py
+++ b/tests/test_attentions.py
@@ -96,9 +96,7 @@ def test_order_invariance(
     )
 
     # Check that a shuffled input produces the same results
-    seqs = (
-        [SEQ, SEQ - 16] if (attention_name != "blocksparse" and not causal) else [SEQ]
-    )
+    seqs = [SEQ, SEQ - 16] if (attention_name != "blocksparse") else [SEQ]
 
     for seq in seqs:
         # Check that we can pass a smaller sequence
@@ -189,9 +187,9 @@ def test_different_kq_dimensions(
 
 
 @pytest.mark.parametrize("heads", [1, 4])
-@pytest.mark.parametrize("attention_name", ["scaled_dot_product"])
-@pytest.mark.skipif(torch.cuda.is_available(), reason="requires a CUDA gpu")
-def test_scaled_dot_product_causal(
+@pytest.mark.parametrize("attention_name", ["scaled_dot_product", "favor"])
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="requires a CUDA gpu")
+def test_causal(
     attention_name: str,
     heads: int,
 ):
@@ -218,7 +216,7 @@ def test_scaled_dot_product_causal(
         .expand(1, -1, -1)
     )
     q = (
-        torch.triu(torch.ones((SEQ, SEQ), device=device), diagonal=1)
+        torch.triu(torch.ones((SEQ, SEQ), device=device), diagonal=0)
         .unsqueeze(0)
         .expand(1, -1, -1)
     )
--- a/tests/test_attentions.py
+++ b/tests/test_attentions.py
@@ -198,6 +198,8 @@ def test_causal(
     The input data is orthogonal by design if causal is respected, but if the attention looks ahead this will fail
     """
 
+    torch.random.manual_seed(42)
+
     device = torch.device("cuda")
 
     multi_head = _get_multihead(
@@ -234,9 +236,10 @@ def test_causal(
     res = multi_head(query=q, key=k, value=v).squeeze(0)
 
     # Consolidate along the embedding, if causal was respected the amplitude should be sorted already
-    res_sum = torch.sum(res, dim=1)
-    assert torch.allclose(torch.sort(res_sum)[0], res_sum) or torch.allclose(
-        torch.sort(res_sum, descending=True)[0], res_sum
+    res_sum = torch.sum(res, dim=1).cpu()
+
+    assert torch.allclose(torch.sort(res_sum)[1], torch.arange(SEQ)) or torch.allclose(
+        torch.sort(res_sum, descending=True)[1], torch.arange(SEQ)
     ), res_sum
 
 
--- a/tests/test_attentions.py
+++ b/tests/test_attentions.py
@@ -29,7 +29,15 @@ GLOBAL_ATTENTION_RATIO = (
 assert ATTENTION_REGISTRY.keys(), "Attention layers should have been registered"
 
 
-def _get_multihead(attention_name, attn_dropout, res_dropout, causal, heads, device):
+def _get_multihead(
+    attention_name,
+    attn_dropout,
+    res_dropout,
+    causal,
+    heads,
+    device,
+    skip_output_projection=False,
+):
     test_config = {
         "name": attention_name,
         "dropout": attn_dropout,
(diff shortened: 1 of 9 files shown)


SeanNaren (approved):

blefaudeux:
The test I came up with is a bit strict, so it's effective with scaled_dot_product but it does not pass with all the approximations, they're mostly there but not exactly. I think it would be nice to come up with a proper test for all of them, but in the meantime this needs to be fixed on main

xformers/components/attention/favor.py
```diff
@@ -24,6 +24,7 @@
 
 @dataclass
 class FavorAttentionConfig(AttentionConfig):
+    causal: Optional[bool]
```
blefaudeux, line 27: causal was not being passed on with the factory, because of that..

blefaudeux:
back to draft, looks like favor causal needs improvements, long overdue (spotted by @fmassa long ago)

xformers/components/attention/favor.py
```diff
@@ -109,33 +110,28 @@ def _maybe_promote(x: torch.Tensor) -> torch.Tensor:
     def _causal_attention(
         k_prime: torch.Tensor, q_prime: torch.Tensor, v: torch.Tensor
     ) -> Tuple[torch.Tensor, torch.Tensor]:
-        # TODO(@lefaudeux): Rewrite as suggested in
-        # https://github.com/fairinternal/xformers/pull/61#discussion_r628420093
         # Algorithm 1 in the paper
-        ref_v = torch.ones_like(v[:, 0, :].unsqueeze(1))
+        ref_v = torch.ones_like(v.unsqueeze(2))  # BATCH x SEQ x 1 x EMB
```
blefaudeux, line 114: this is a memory hog, that's the issue with this approach @fmassa, but the older state was not actually trainable (this was not being tested.. fixed now), so I guess that it's better than before and improving on that could be a follow up

blefaudeux: see https://github.com/facebookresearch/xformers/issues/105 for tracking

blefaudeux:
@dianaml0 breaking now around Nystrom + key padding mask, I'm seeing a few lines that I don´t get (like mask and key_padding_mask being used interchangeably ?)

dianaml0:
> @dianaml0 breaking now around Nystrom + key padding mask, I'm seeing a few lines that I don´t get (like mask and key_padding_mask being used interchangeably ?)

Taking a look now! For Nystrom only `key_padding_mask` is accepted. And if the causal flag is specified, `key_padding_mask` is combined with a causal mask into `mask`. I'll check the errors

xformers/components/attention/nystrom.py
```diff
@@ -194,8 +194,9 @@ def forward(
             mask: Optional[torch.Tensor] = None
 
             if self.causal:
-                mask = self._tril_mask(batched_dim, seq_len, seq_len, **tt)
+                mask = self._triu_mask(batched_dim, seq_len, seq_len, **tt)
 
+            # FIXME: @lefaudeux This is probably broken, we don´t test this I suppose
```
dianaml0, line 199: This should be tested in `test_masking` in `test_nystrom_attention.py`

tests/test_nystrom_attention.py
```diff
@@ -85,8 +85,10 @@ def test_att_mask_ignored():
         assert torch.allclose(r_nystrom, r_sdp, rtol=0.005, atol=1e-2)
 
     def test_masking():
-        nystrom_config["causal"] = True
-        sdp_config["causal"] = True
+        # FIXME: Masking seems fine as such, but whole lines are randomly masked when causal+random mask
```
dianaml0, line 88: Is this what you meant by key padding + causal being broken?

xformers/components/attention/nystrom.py
blefaudeux: it's this line which I thought could not possibly work, given the difference in shapes ?

tests/test_nystrom_attention.py
blefaudeux: yes, although it feels like it could be only an issue with the unit test, but scaled_dot_product and nystrom are not behaving in the same way (although there are plenty of NaNs around in both cases). I think that it could be that Nystrom breaks earlier because of the pooling for anchors, if one anchor is full of inf (easy with causal) it cannot compute the attention

blefaudeux: Fixed now with a random masking a little less aggressive (not masking half of the words :)), I think that it's still a good test

xformers/components/attention/nystrom.py
dianaml0: I think it should work since we assert on `key_padding_mask` dimension being (batch, 1, seq)? And we also convert it to an additive mask. But perhaps using `masked_fill` with a boolean key padding mask makes more sense?

xformers/components/attention/scaled_dot_product.py
```diff
@@ -74,24 +74,40 @@ def forward(
                     - If the mask has the float type, then an additive mask is expected (masked values are -inf)
 
         """
+
+        # Handle a possibly deferred causal mask handling
```
dianaml0, line 78: Nice catching of edge cases!

dianaml0 (approved):

tests/test_nystrom_attention.py
dianaml0: Nice :) So can causal be used for the test now or that's still an issue?

tests/test_attentions.py
```diff
@@ -81,7 +96,9 @@ def test_order_invariance(
     )
 
     # Check that a shuffled input produces the same results
-    seqs = [SEQ, SEQ - 16] if attention_name != "blocksparse" else [SEQ]
+    seqs = (
+        [SEQ, SEQ - 16] if (attention_name != "blocksparse" and not causal) else [SEQ]
```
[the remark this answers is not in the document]
blefaudeux: ah, good question ! Basically we adjust the mask on the fly when the sequence is smaller than initially thought, but we don't have the codepath in place to also adjust the causal mask so that broke (it "worked" before because we were not actually applying the causal mask..). I can fix that in this PR and then refactor all this part in a sequel ?

tests/test_nystrom_attention.py
blefaudeux: not an issue anymore, I messed up the push when rebasing but it works fine locally, let me update..

xformers/components/attention/nystrom.py
blefaudeux: yes you're right, sorry I forgot again, the 1 will be enough to broadcast ! all good, 'll remove the FIXME, I've changed the unit test to make sure that we do go there and it works indeed

tests/test_attentions.py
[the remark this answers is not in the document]
dianaml0: Ah okay thanks for the explanation :) That sounds good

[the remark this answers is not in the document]
dianaml0: Ah okay thanks for the explanation :) That sounds good

PR #111, opened 2021-11-17
[deploy] Updating the setup, still not ideal

blefaudeux:
## What does this PR do?
Minor, but make sure that installing from the pip package does not try to build the extension (files are not there). Checking with @fmassa, the idea here is that people don't have the compilers anyway, another option would be to ship the files and improve in the setup so that if no way to build the extensions this does not fail. 

Tested on two machines, the pip package installation now works nicely, but this definitely feels imperfect

## Before submitting

- [ ] Did you have fun?
- [ ] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/master/CONTRIBUTING.md)?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ There are two ways you can install it:
   ```bash
   conda create --name xformer_env
   conda activate xformer_env
+  pip install torch
   pip install xformers
   ```
 
--- a/setup.py
+++ b/setup.py
@@ -89,29 +89,34 @@ def get_extensions():
 
     sources = [os.path.join(extensions_dir, s) for s in sources]
 
-    ext_modules = [
-        extension(
-            "xformers._C",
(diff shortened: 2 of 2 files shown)


fmassa (commented):
Hum, I would have included the C++ / CU files in the wheels.
Probably need to add `package_data` in the `setup` command, like we do in https://github.com/pytorch/vision/blob/aedd39792d07af58e55ec028ed344d63debbd281/setup.py#L467

blefaudeux:
dropping this, discussing with @fmassa and I agree, better do the right thing. As a stopgap the current xformers package works for everyone (although without the sources), and we'll fix the package and release process properly for the next version

Issue #91, opened 2021-11-09
`scaled_dot_product_attention` shouldn't have a `causal` flag


fmassa:
The causal flag in https://github.com/facebookresearch/xformers/blob/cc7c8edd6907abdb09d10f60e456e224109a4aca/xformers/components/attention/core.py#L191 was added to enable fast triton kernels for softmax.

I believe we shouldn't have this flag, as it makes it look like we can combine `attn_mask` and `causal`, which doesn't look like it's the case.
Plus, the behavior of `causal` is embedded inside `attn_mask` already.

IMO we should remove `causal`, and instead have a custom class `UpperTriangularMatrix` which can be passed as an `attn_mask`, which gets then dispatched to the efficient causal softmax

blefaudeux:
ah interesting, I went over this part a couple of times recently (with the additive/bool mask change) and it did feel a bit messy. Ok with the proposed solution, in general I think that having types around the mask would help cornering the usecases, beyond causal (ie: convey the fact that it's additive)

PR #113, opened 2021-11-18
[partial refactor] Dedicated attention mask wrapper

blefaudeux:
## What does this PR do?
- add a dedicated class to handle additive attention masks, and centralize some of the typical operations (change types, add to another mask, handle smaller sequences..)
- the sparse machinery is still the same and bools (cc @fmassa), I did not touch that. 
- I think that the above needs to be changed eventually, so that we converge towards additive masks all around, but it means that on the sparse side we need to completely handle a float additive mask, in that a value `a` would both mark the fact that this needs to be computed, and be added a posteriori to the computed attention value (not done right now)
- Fix https://github.com/facebookresearch/xformers/issues/90 (do not slice the attention mask)
- Fix https://github.com/facebookresearch/xformers/issues/91 (no causal flag in the core attention)

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -35,7 +35,7 @@ copyright = "2021, Facebook AI Research"
 author = "Facebook AI Research"
 
 # The full version, including alpha/beta/rc tags
-release = "0.0.5"
+release = "0.0.6"
 
 
 # -- General configuration ---------------------------------------------------
--- a/tests/test_pytorch_transformer_parity.py
+++ b/tests/test_pytorch_transformer_parity.py
@@ -114,7 +114,6 @@ def test_pytorch_encoder_parity(device=torch.device("cuda")):
             dim_feedforward=4 * EMB,
             dropout=DROP,
             activation=ACTIVATION,
-            layer_norm_eps=1e-05,
             batch_first=True,  # (batch, seq, feature)
             device=device,
         ),
--- a/xformers/components/attention/__init__.py
+++ b/xformers/components/attention/__init__.py
@@ -16,6 +16,7 @@ from xformers.utils import (
 )
 
 from ._sputnik_sparse import SparseCS
+from .attention_mask import AttentionMask
 from .base import Attention, AttentionConfig  # noqa
 
 # CREDITS: Classy Vision registry mechanism
@@ -77,10 +78,11 @@ register_attention: Callable[[str, Any], Callable[[Any], Any]] = get_registry_de
 )
 
 
-def maybe_sparsify(matrix):
+def maybe_sparsify(matrix) -> Any:
     # Sparsify if that makes sense
     if torch.count_nonzero(matrix).item() / matrix.numel() > _DENSITY_THRESHOLD:
-        return matrix
+        # If not sparse, then AttentionMask is the reference type
+        return AttentionMask.from_bool(matrix)
 
     return sparsify(matrix)
 
@@ -110,6 +112,7 @@ __all__ = [
     "GlobalAttention",
     "FavorAttention",
     "Attention",
+    "AttentionMask",
     "build_attention",
     "register_attention",
 ]
--- a/xformers/components/attention/attention_mask.py
+++ b/xformers/components/attention/attention_mask.py
@@ -1,0 +1,144 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+from typing import Optional, Type, TypeVar
+
+import torch
+
+Self = TypeVar("Self", bound="AttentionMask")
+
+
+class AttentionMask:
+    """
+    Holds an attention mask, along with a couple of helpers and attributes.
+
+    .. note: this is an additive mask, meaning that coefficients which should be computed hold the '0.' value,
+        and coefficients which should be skipped hold the '-inf' value. Any other value is possible if the purpose
+        is to bias the attention computation for instance
+
+    .. note: the attention mask dimensions are expected to be `[batch, to_sequence, from_sequence]`,
+        `[to_sequence, from_sequence]`, or anything broadcastable in between
+    """
+
+    def __init__(self, additive_mask: torch.Tensor, is_causal: bool = False):
+        assert additive_mask.is_floating_point()
+        assert not additive_mask.requires_grad
+
+        if additive_mask.ndim == 2:
+            additive_mask = additive_mask.unsqueeze(0)
+
+        self.values = additive_mask
+        self.is_causal = is_causal
+        self.seq_len = additive_mask.shape[1]
+        self.to_seq_len = additive_mask.shape[0]
+
+    def to_bool(self) -> torch.Tensor:
+        """
+        .. warning: we assume here that True implies that the value should be computed
+        """
+        return self.values != float("-inf")
+
+    @classmethod
+    def from_bool(cls: Type[Self], x: torch.Tensor) -> Self:
+        """
+        Create an AttentionMask given a boolean pattern.
+        .. warning: we assume here that True implies that the value should be computed
+        """
+        assert x.dtype == torch.bool
+
+        additive_mask = torch.empty_like(x, dtype=torch.float)
+        additive_mask.masked_fill_(x, 0.0)
+        additive_mask.masked_fill_(~x, float("-inf"))
+
+        return cls(additive_mask)
+
+    @classmethod
+    def from_multiplicative(cls: Type[Self], x: torch.Tensor) -> Self:
+        """
+        Create an AttentionMask given a multiplicative attention mask.
+        """
+        assert not x.dtype == torch.bool
+
+        additive_mask = torch.empty_like(x, dtype=torch.float)
+        x = x.bool()
+
+        additive_mask.masked_fill_(x, 0.0)
+        additive_mask.masked_fill_(~x, float("-inf"))
+
+        return cls(additive_mask)
+
+    @classmethod
+    def make_causal(
+        cls: Type[Self],
+        seq_len: int,
+        to_seq_len: Optional[int] = None,
+        device: Optional[torch.device] = None,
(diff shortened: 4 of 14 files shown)


docs/source/conf.py
```diff
@@ -35,7 +35,7 @@
 author = "Facebook AI Research"
 
 # The full version, including alpha/beta/rc tags
-release = "0.0.5"
+release = "0.0.6"
```
blefaudeux, line 38: just bumped to mark a difference with the released package (0.0.5)

tests/test_pytorch_transformer_parity.py
```diff
@@ -114,7 +114,6 @@ def test_pytorch_encoder_parity(device=torch.device("cuda")):
             dim_feedforward=4 * EMB,
             dropout=DROP,
             activation=ACTIVATION,
-            layer_norm_eps=1e-05,
```
blefaudeux, line 117: not present in some pytorch versions, and this is the default anyway

xformers/components/attention/__init__.py
```diff
@@ -77,10 +78,11 @@ class MyAttention(Attention):
 )
 
 
-def maybe_sparsify(matrix):
+def maybe_sparsify(matrix) -> Any:
     # Sparsify if that makes sense
     if torch.count_nonzero(matrix).item() / matrix.numel() > _DENSITY_THRESHOLD:
-        return matrix
+        # If not sparse, then AttentionMask is the reference type
```
blefaudeux, line 84: heads up on that, basically hijacking any non-sparse case to put it under the AttentionMask umbrella

xformers/components/attention/attention_mask.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+from typing import Optional, Type, TypeVar
+
+import torch
+
+Self = TypeVar("Self", bound="AttentionMask")
+
+
+class AttentionMask:
```
blefaudeux, line 14: tentatively own all attention masks, but this "breaks" (not per say, but not handled here) right now when things are sparsified

xformers/components/attention/core.py
```diff
@@ -69,12 +70,12 @@ def _broadcast_batch(mask, batch_size):
 
 
 def _matmul_with_mask(
-    a: torch.Tensor, b: torch.Tensor, mask: Optional[torch.Tensor]
+    a: torch.Tensor, b: torch.Tensor, mask: Optional[Union[torch.Tensor, "SparseCS"]]
```
blefaudeux, line 73: that's where the seams break a little in between AttentionMask and the sparse takes. On the bright side, it's all internal. On the dark side, ideally we could have something a little more streamlined overall.

xformers/components/attention/core.py
```diff
@@ -199,25 +201,26 @@ def scaled_query_key_softmax(
...
     # Matmul with mask, if boolean
-    is_bool_mask = att_mask is not None and att_mask.dtype == torch.bool
-    att = _matmul_with_mask(q, k.transpose(-2, -1), att_mask if is_bool_mask else None)
+    if att_mask is not None and isinstance(att_mask, AttentionMask):
+        # Additive mask
+        mask: Optional[Union[SparseCS, torch.Tensor]] = att_mask.values
+    else:
+        mask = att_mask
 
-    # Could also be that the mask was additive
```
blefaudeux, line 205: @dianaml0 _matmul_with_mask already has an additive mask codepath actually. It was written twice, but the computation was correct (since we only passed the mask to _matmul_with_mask if it was a bool..). Maybe that we should keep the addition here actually and remove it from matmul ?

xformers/components/attention/core.py
```diff
@@ -199,25 +201,26 @@ def scaled_query_key_softmax(
...
-    if att_mask is not None and att_mask.dtype != torch.bool:
-        att = att + att_mask
+    att = _matmul_with_mask(q, k.transpose(-2, -1), mask)
 
     # Softmax to get the attention probabilities
-    att = _softmax(att, causal=causal)
+    is_causal = att_mask.is_causal if isinstance(att_mask, AttentionMask) else False
```
blefaudeux, line 213: this fixes https://github.com/facebookresearch/xformers/issues/91

xformers/components/attention/core.py
```diff
@@ -231,30 +234,23 @@ def scaled_dot_product_attention(
...
-            if not att_mask.is_sparse:
-                att_mask = att_mask[:seq, :seq]
-            else:
-                logging.warning(
-                    "Mismatching attention mask and sequence length. On the fly correction but this will be slow"
-                )
-                # Loosing sparsity on purpose,
-                # expectation is that moving back and forth dense/sparse will negate the speedup
-                att_mask = att_mask.to_dense().squeeze(0)[:seq, :seq]
+        if isinstance(att_mask, AttentionMask):
```
blefaudeux, line 237: this is cleaner than before I believe, but the dense/sparse case is exposed

xformers/components/attention/global_tokens.py
```diff
@@ -91,10 +92,24 @@ def forward(
             self.attention_mask = self.attention_mask.to(q.device)
 
         # Mask-aware attention
-        mask = (
-            self.attention_mask if att_mask is None else self.attention_mask & att_mask
+        if att_mask is not None:
```
blefaudeux, line 95: same, sparse/dense handling discrepancy, fixing this would be a nice follow up

xformers/components/attention/random.py
```diff
@@ -93,12 +94,24 @@ def forward(
             self.rand_attention_mask = self._get_rand_mask(q.shape).to(q.device)
 
         # Mask-aware attention
-        mask = (
-            self.rand_attention_mask
-            if att_mask is None
-            else self.rand_attention_mask & att_mask
+        if att_mask is not None:
```
blefaudeux, line 97: same, sparse/dense discrepancy

xformers/components/attention/attention_mask.py
```diff
@@ -0,0 +1,144 @@
...
+        self.to_seq_len = additive_mask.shape[0]
+
+    def to_bool(self) -> torch.Tensor:
+        """
+        .. warning: we assume here that True implies that the value should be computed
+        """
+        return self.values != float("-inf")
+
+    @classmethod
+    def from_bool(cls: Type[Self], x: torch.Tensor) -> Self:
```
dianaml0, line 45: Really nice to have these!

dianaml0 (approved):
Really nice to have this all centralized!

xformers/components/attention/core.py
fmassa: Thanks!

One downside of this approach is that it leaks internal implementation details of `AttentionMask` inside a core function.

My thoughts were to modify instead `_softmax` so that the causal path is just a different function call from a `CausalAttentionMask` Tensor type.

xformers/components/attention/core.py
fmassa: I still believe we should not be cropping the mask in our core functions. But ok to keep this as is for now

fmassa (commented):
Thanks for the work @blefaudeux !

I had some slightly different thoughts about the implementation, but let's move with what you have for now. I think it might be worth getting in a call and discuss.

I propose we get this merged, but IMO we need to re-write this whole file to entirely take into account the things we care about and get the abstractions that best suit our new use-cases.

xformers/components/attention/core.py
blefaudeux: in that case you pass the mask to softmax, it moves the branching there, why not. Typically softmax do not take masks though, so I think that the interface becomes a little confusing (we'll pass a mask in only to say that it's causal or not)

xformers/components/attention/core.py
blefaudeux: it's also a matter of deduplicating, else it's something which has to be done in all the corresponding attentions. No strong opinion though, I'm ok to move that up in SDP and others

xformers/components/attention/core.py
```diff
@@ -69,12 +70,12 @@ def _broadcast_batch(mask, batch_size):
...
 
 def _matmul_with_mask(
-    a: torch.Tensor, b: torch.Tensor, mask: Optional[torch.Tensor]
+    a: torch.Tensor, b: torch.Tensor, mask: Optional[Union[torch.Tensor, "SparseCS"]]
 ) -> torch.Tensor:
     if mask is None:
         return a @ b
 
-    if _is_sparse_available:
+    if _is_sparse_available and mask.dtype == torch.bool:
```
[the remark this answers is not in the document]
blefaudeux: ideally I think that we should just expose one mask type, it's easier to understand for the user, and do the appropriate thing internally from there (sparse, dense, blocksparse, I agree with @fmassa on that). I think that it's beyond this PR though, so at that time yes if you pass a bool mask it goes the maybe_sparse way, if you pass a float then it's all dense. I can change that, this is a draft !

xformers/components/attention/core.py
blefaudeux: fixed that @fmassa :)

xformers/components/attention/core.py
blefaudeux: @fmassa is that ok if you propose a change (object oriented if I got it right, the masks handle the dispatch problem) in another PR ?

xformers/components/attention/core.py
```diff
@@ -199,25 +201,26 @@ def scaled_query_key_softmax(
     q = q / math.sqrt(k.size(-1))
 
     # Matmul with mask, if boolean
```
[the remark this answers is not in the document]
blefaudeux: should be fixed

blefaudeux:
landing that as I understood that @dianaml0 and @fmassa were ok with this as a first take, but @fmassa to propose a rewrite where attention mask(s) handle the specific computations redirect (personally ok with that)

xformers/components/attention/core.py
fmassa: Sure, I'll send a PR with my proposed changes later this week

Issue #108, opened 2021-11-16
Benchmark with Timm


blefaudeux:
# 🚀 Feature
[Timm](https://github.com/rwightman/pytorch-image-models/blob/master/timm/models/vision_transformer.py#L177) is the SOTA in the field for vision transformers, it would be a good touchstone to check where a vanilla xformers speed lands

## Motivation
speed is important to be compelling

PR #119, opened 2021-11-20
[fix][speed] Better projections - correctness + speed

blefaudeux:
## What does this PR do?
- fixes the self attention optimization (project all in one go) not being triggered
- contiguous tensors are only needed for sparse, removing the contiguous cost when not needed
- fixing a simplifying a broken path in the factory / residual and normalization, related to the above: only normalize once if this is the same tensor, and propagate the same id down

With these changes/fixes, vanilla xformers is 5-10% faster than timm on a vanilla ViT (up from being 10% slower) as per the bench [from the other PR](https://github.com/facebookresearch/xformers/issues/108), and microGPT trains something decent in 20 minutes on a laptop (3080 / fp16)

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/xformers/components/attention/_sputnik_sparse.py
+++ b/xformers/components/attention/_sputnik_sparse.py
@@ -258,7 +258,7 @@ class SparseCS:
         column_indices = self.column_indices
         out = _sddmm.apply(
             a,
-            b.transpose(-2, -1),
+            b.transpose(-2, -1).contiguous(),
             row_indices,
             row_offsets,
             column_indices,
--- a/xformers/components/in_proj_container.py
+++ b/xformers/components/in_proj_container.py
@@ -153,17 +153,14 @@ class InProjContainer(nn.Module):
         if self.in_proj_weight is not None:
             if id(query) == id(key):
                 # Self attention, get all the projected values at once
-                # we compute everything transposed, so that q,k,v stay contiguous after splitting
+                # NOTE: the resulting buffers could be contiguous out of the box with a custom kernel
                 qkv = query @ self.in_proj_weight.transpose(-2, -1)
 
                 if self.in_proj_bias is not None:
                     qkv += self.in_proj_bias
 
-                q, k, v = map(
-                    lambda x: x.contiguous(),
-                    qkv.split(self.out_features, dim=-1),
-                )
-                return q, k, v
+                qkv = qkv.split(self.out_features, -1)
+                return qkv[0], qkv[1], qkv[2]
 
             else:
                 # Not self attention
--- a/xformers/components/residual.py
+++ b/xformers/components/residual.py
@@ -5,23 +5,14 @@
 
 
 from enum import Enum
-from typing import List, Union
+from typing import Union
 
-import torch
 import torch.nn as nn
 
 from xformers import _is_triton_available
 
 if _is_triton_available:
     from xformers.triton.layer_norm import FusedLayerNorm
-
-
-def _to_tensor_list(
-    inputs: Union[torch.Tensor, List[torch.Tensor]]
-) -> List[torch.Tensor]:
-    if not isinstance(inputs, list):
-        inputs = [inputs]
-    return inputs
 
 
 class LayerNormStyle(str, Enum):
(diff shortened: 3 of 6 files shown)


xformers/components/attention/_sputnik_sparse.py
```diff
@@ -258,7 +258,7 @@ def matmul_with_mask(self, a, b):
         column_indices = self.column_indices
         out = _sddmm.apply(
             a,
-            b.transpose(-2, -1),
+            b.transpose(-2, -1).contiguous(),
```
blefaudeux, line 261: contiguous is only needed here and down, we were forcing this all the time before. Note that with a good projection kernel this could come for free, could be nice

xformers/components/in_proj_container.py
```diff
@@ -153,17 +153,14 @@ def forward(
...
 
                 if self.in_proj_bias is not None:
                     qkv += self.in_proj_bias
 
-                q, k, v = map(
-                    lambda x: x.contiguous(),
-                    qkv.split(self.out_features, dim=-1),
-                )
-                return q, k, v
+                qkv = qkv.split(self.out_features, -1)
```
blefaudeux, line 162: this was actually slow (memcopy) and not needed for non-sparse attention

xformers/components/residual.py
```diff
@@ -36,16 +27,23 @@ class LayerNormStyle(str, Enum):
...
+
+    .. warning: by convention, if multiple tensors are being passed in,
+        the first one is used for the residual path
+    """
 
     def __init__(self, layer: nn.Module):
         super().__init__()
         self.layer = layer
 
-    def forward(self, inputs: Union[torch.Tensor, List[torch.Tensor]], *args, **kwargs):
```
blefaudeux, line 45: this was super error prone (@stephenroller spotted that long ago, I should have caught that then), since intputs and args can mix

xformers/components/residual.py
```diff
@@ -62,11 +60,16 @@ def __init__(self, d_model: int, sublayer: nn.Module, use_triton: bool = True):
...
-        inputs = _to_tensor_list(inputs)
-
-        x_norm = [self.norm(x_) for x_ in inputs]
-        return self.sublayer(*x_norm, *args, **kwargs)
+    def forward(self, *args, **kwargs):
+        # Could be that the same tensor has been passed multiple times
+        # in that case we'll just normalize once
+        list_ids = [id(inp) for inp in args]
```
blefaudeux, line 66: seems small, but here if the same tensor was actually passed multiple times (self attention), we would normalize 3 times and loose the same id(), which in turn means that in the attention layer we would not optimize for self-attention

xformers/factory/block_factory.py
```diff
@@ -364,8 +364,8 @@ def forward(
         else:
             target_q, target_k, target_v = target, target, target
 
-        x = self.wrap_att([target_q, target_k, target_v], att_mask=decoder_att_mask)
-        x = self.wrap_cross([x, memory, memory], att_mask=encoder_att_mask)
+        x = self.wrap_att(target_q, target_k, target_v, att_mask=decoder_att_mask)
```
blefaudeux, line 367: related to the "input<>args" cleanup

blefaudeux:
@fmassa @dianaml0 I think that the code in block_factory and residual is probably way to complicated and error prone, could be worth another pass if deemed important (this is orthogonal to the parts zoo approach). In any case with these changes a default xformers is competitive in terms of speed vs. Timm ViT

blefaudeux:
Ping reviewers (and PR below), this significantly speeds up a code path used in the examples and by some folks around

dianaml0:
> @fmassa @dianaml0 I think that the code in block_factory and residual is probably way to complicated and error prone, could be worth another pass if deemed important (this is orthogonal to the parts zoo approach). In any case with these changes a default xformers is competitive in terms of speed vs. Timm ViT

Seems like much of the complication was introduced for reversible layers to work? But agree probably a good idea to see if there's a cleaner way to do it.

dianaml0 (approved):
This is really great!! LGTM!

blefaudeux:
> > @fmassa @dianaml0 I think that the code in block_factory and residual is probably way to complicated and error prone, could be worth another pass if deemed important (this is orthogonal to the parts zoo approach). In any case with these changes a default xformers is competitive in terms of speed vs. Timm ViT
> 
> 
> 
> Seems like much of the complication was introduced for reversible layers to work? But agree probably a good idea to see if there's a cleaner way to do it.

Yes for reversible, I agree, it makes the signal paths a lot more complex unfortunately:( no strong opinion on that, I don't think that this was benchmarked enough to have a definitive answer. Probably possible to have a better code and keep reversible though

xformers/components/attention/_sputnik_sparse.py
stephenroller: maybe you should leave a comment/breadcrumb trail saying that in the code?

blefaudeux: will do, or maybe just file an issue, basically the projection in the beginning of MHA could be hardened a little for speed

PR #123, opened 2021-11-28
[Fix] attn_dropout

### LINKED ISSUE #122: `BlockSparseAttention` doesn't use dropout
# 🐛 Bug

While working on refactoring the `blocksparse.py` file into its own custom tensor type, I noticed that `BlockSparseAttention` doesn't implement dropout, which is probably not intended.

cc @blefaudeux

blefaudeux:
## What does this PR do?
Fixes #122 + adds a unit test to catch faulty dropouts
@fmassa is rewriting some of that part, but in the meantime `main` should not be knowingly broken

## Before submitting

- [x] Did you have fun?
- [ ] Did you read the [contributor guideline](https://github.com/facebookresearch/fairscale/blob/master/CONTRIBUTING.md)?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/tests/test_attentions.py
+++ b/tests/test_attentions.py
@@ -77,7 +77,7 @@ def _get_multihead(
     return multi_head
 
 
-@pytest.mark.parametrize("attn_dropout", [0.0, 0.9])
+@pytest.mark.parametrize("attn_dropout", [0.0, 0.3])
 @pytest.mark.parametrize("residual_dropout", [0.0, 0.1])
 @pytest.mark.parametrize("causal", [True, False])
 @pytest.mark.parametrize("heads", [1, 4])
@@ -91,6 +91,9 @@ def test_order_invariance(
     causal: bool,
     device: torch.device,
 ):
+
+    torch.manual_seed(42)
+
     multi_head = _get_multihead(
         attention_name, attn_dropout, residual_dropout, causal, heads, device
     )
--- a/tests/test_attentions.py
+++ b/tests/test_attentions.py
@@ -77,7 +77,7 @@ def _get_multihead(
     return multi_head
 
 
-@pytest.mark.parametrize("attn_dropout", [0.0, 0.5])
+@pytest.mark.parametrize("attn_dropout", [0.0, 0.9])
 @pytest.mark.parametrize("residual_dropout", [0.0, 0.1])
 @pytest.mark.parametrize("causal", [True, False])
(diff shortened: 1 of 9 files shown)


blefaudeux:
@fmassa @dianaml0 turns out that in many case dropout was not applied, not just for blocksparse, a unit test is catching that now :(

dianaml0:
> @fmassa @dianaml0 turns out that in many case dropout was not applied, not just for blocksparse, a unit test is catching that now :(

Oh wow :( Great that you were able to catch that!

blefaudeux:
> > @fmassa @dianaml0 turns out that in many case dropout was not applied, not just for blocksparse, a unit test is catching that now :(
> 
> Oh wow :( Great that you were able to catch that!

maybe a semi-recent change, I think that @fmassa is right in that we should refactor core/. Else for some variant the setting was silently ignored (but not explicitly present in the constructor), so it would just fail for the factory, not as bad but still... Unit tests for the win, this should not happen again

xformers/components/attention/core.py
```diff
@@ -172,10 +173,14 @@ def _apply_dropout(att, dropout):
...
             values = dropout(values)
             att = torch.sparse_coo_tensor(att.indices(), values, att.shape)
+        else:
+            # Simple dense case
+            att = dropout(att)
 
         return att
 
     # Non optimized vanilla dropout
+    print(att - dropout(att))
```
dianaml0, line 183: Should this be removed?

tests/test_attentions.py
```diff
@@ -110,7 +113,12 @@ def test_order_invariance(
         torch.allclose(results[:, shuffle, :], results_shuffled)
 
         # Test the non-self-attention codepath
-        _ = multi_head(inputs, inputs_shuffled, inputs)
+        att = multi_head(inputs, inputs_shuffled, inputs)
+
+        # Check that dropout actually drops some values
```
dianaml0, line 118: Great to have this test now!

xformers/components/attention/linformer.py
```diff
@@ -37,6 +37,7 @@ def __init__(
         if k is None:
             k = seq_len // 4
 
+        print("dropout ", dropout)
```
dianaml0, line 40: Same here?

dianaml0 (approved):
Really nice catch!

PR #124, opened 2021-11-29
Introduce SparseCSRTensor as a replacement for SparseCS

fmassa:
# SparseCSRTensor

This PR adds a custom Tensor class `SparseCSRTensor`, which inherits from `torch.Tensor`, and performs all the operators that were previously implemented in `SparseCS` in `SparseCSRTensor`.

The benefits of having `SparseCSRTensor` being a Tensor subclass are the following:
- they can be used as a `buffer` in a `nn.Module`, so that casting a model to `.to('cuda')` also casts the `SparseCSRTensor`. They are also present in `state_dict`
-  this is a first step towards enabling `SparseCSRTensor` to be a parameter of a module (it still requires changes to PyTorch for it to work though, see https://github.com/pytorch/pytorch/issues/65018)
- we have a `.grad` field that returns a `SparseCSRTensor`

Additionally, there are the following benefits of using `__torch_function__` in `SparseCSRTensor`:
- we can use interchangeably any torch function with `torch.Tensor` as well as `SparseCSRTensor` (as long as the operator has been implemented)
- we let the PyTorch dispatcher to handle the dispatch for us

# masked_matmul

In addition to `SparseCSRTensor`, this PR also introduces a `masked_matmul` function that performs the same operations as `matmul_with_mask` from before. The difference is that it adds the necessary mechanisms for `__torch_function__` to work, and thus is generic and can be used with `SparseCSRTensor` as well.

Note that I plan to rename / refactor this function by removing the `bool` support and making it perform simply `mat1 @ mat2 + bias`, and I'll probably rename it `add_matmul`, similarly to `addmm` (but which only supports 2d matrices).

# Examples

Here are a few examples of how the new `SparseCSRTensor` and `masked_matmul` can be used:
```python
import torch
from xformers.ops import masked_matmul
from xformers.sparse import SparseCSRTensor

matrix = torch.rand(2, 3, 3)
matrix[(matrix[0:1] < 0.5).repeat(2, 1, 1)] = 0
matrix *= 10

mat = SparseCSRTensor.from_dense(matrix)

# shows that it works with `nn.Module`
module = torch.nn.Module()
module.register_buffer("a", mat)
s = module.state_dict()
assert s['a'] is mat
module.load_state_dict(s)

# operations work
torch.matmul(mat, matrix)
torch.nn.functional.softmax(mat, -1)
torch.nn.functional.dropout(mat, 0.5)
masked_matmul(matrix, matrix.transpose(-2, -1), mat)
# plus a few other operators
```

# PyTorch 1.11+

One drawback of this approach is that it relies on features from PyTorch nightly, which are not yet in a stable release.
This can be an annoyance for some, but we can cut a new release for `xformers` with this new requirement

# Looking ahead

This PR is meant to be entirely BC-compatible wrt `SparseCS` and the other operators. This minimizes code changes and simplifies review.

## Cleanup `core.py`

The next step is to clean up `core.py` to rely on the `SparseCSRTensor`. Thanks to the PyTorch dispatch we are using, I expect it to look much simpler than it currently is. Indeed, the implementations would look something like
```python
def scaled_query_key_softmax(q, k, att_mask):
    q = q / math.sqrt(k.size(-1))
    att = masked_matmul(q, k.transpose(-2, -1), att_mask)
    # note that causal is just a different class of matrix that will dispatch
    # to the efficient softmax function, and the rest will be just as a standard
    # torch.Tensor
    att = torch.nn.functional.softmax(att, -1)
    return att

def scaled_dot_product_attention(q, k, v, att_mask, dropout):
    att = scaled_query_key_softmax(q, k, att_mask=att_mask)
    att = dropout(att)
    y = att @ v
    return y
```

## Cleanup blocksparse.py

Following the same approach, we can actually remove `blocksparse.py` and make it dispatch to the same core abstractions that we already have. All we need to do is to create a custom `BlockSparseTensor` class that implements the required operators. I've started working on it in [this branch](https://github.com/facebookresearch/xformers/commit/418b9e81c4d4bdbf2fc717e87bca074e8eb1a91e), and plan to submit a PR once this is merged.
The benefits are the same as with `SparseCSRTensor`, plus we can also use block-sparse matrices whenever we call `scaled_dot_product_attention` (including nystromformer etc)


.circleci/config.yml
```diff
@@ -15,12 +15,12 @@ cpu_py38: &cpu_py38
...
-gpu_cu112: &gpu_cu112
+gpu_cu111: &gpu_cu111
   environment:
-    CUDA_VERSION: "11.2"
-    CUDA_HOME: /usr/local/cuda-11.2
+    CUDA_VERSION: "11.1"
+    CUDA_HOME: /usr/local/cuda-11.1
   machine:
-    image: ubuntu-2004-cuda-11.2:202103-01
+    image: ubuntu-1604-cuda-11.1:202012-01
```
fmassa, line 23: Had to change the image to use CUDA 11.1, as PyTorch nightlies now don't support CUDA 11.2 anymore.

requirements-test.txt
```diff
@@ -15,4 +15,4 @@ pytest == 5.4.1
 pytest-cov == 2.10.0
 pytest-mpi == 0.4
 pytest-timeout == 1.4.2
-timm >= 0.4.5
+git+git://github.com/rwightman/pytorch-image-models@v0.4.5#egg=timm
```
fmassa, line 18: The PyPi `timm` package pins to a specific PyTorch / torchvision versions, so I'm installing it from github instead

tests/test_sparse_csr.py
```diff
@@ -0,0 +1,140 @@
...
+    a = a.to_sparse()
+
+    a_csr.requires_grad_(True)
+    fn2(a_csr)._csr_values.sum().backward()
+    grad_a = a_csr._csr_values.grad.clone()
+    a.requires_grad_(True)
+    fn(a).coalesce().values().sum().backward()
+    assert torch.allclose(
+        grad_a, a.grad.coalesce().values().reshape_as(grad_a), atol=1e-7
+    )
```
fmassa, line 97: Those are just tests to validate that the gradients work as expected. They are almost redundant compared to the tests in `test_custom_ops.py`, so I only added one here. Given that `SparseCS` now dispatches entirely to `SparseCSRTensor`, the other tests we have should be enough and thus once I remove `SparseCS` I can move the tests around

tests/test_timm_wrap.py
```diff
@@ -4,15 +4,21 @@
...
-import timm
 import torch
-from timm.models.vision_transformer import VisionTransformer
+
+try:
+    import timm
+    from timm.models.vision_transformer import VisionTransformer
+except ImportError:
+    timm = None
+    VisionTransformer = None
```
fmassa, line 14: Let's not make not having `timm` a hard failure in our tests

xformers/components/attention/_sputnik_sparse.py
```diff
@@ -6,176 +6,11 @@
...
-            values_t,
-            row_offsets_t,
-            column_indices_t,
-        ) = _transpose_with_info(values, _transp_info)
-
-        grad_dense = torch.ops.xformers.spmm_sputnik(
-            grad, row_indices_t, values_t, row_offsets_t, column_indices_t, k
-        )
-
-        return grad_dense, None, grad_sparse, None, None, None, None
```
fmassa, line 178: This has all been moved to `xformers/sparse/_csr_ops.py` without modifications

xformers/components/attention/_sputnik_sparse.py
```diff
@@ -6,176 +6,11 @@
...
-    row_offsets = torch.nn.functional.pad(row_offsets, (1, 0))
-    return row_offsets, column_indices
-
-
-def _csr_to_coo(m, n, row_offsets, column_indices):
-    # convert from compressed rows to uncompressed
-    indices = torch.arange(m, dtype=row_offsets.dtype, device=row_offsets.device)
-    row_sizes = torch.diff(row_offsets)
-    row_coo = torch.repeat_interleave(indices, row_sizes.long())
-    return row_coo, column_indices
```
fmassa, line 22: This has been moved to `xformers/sparse/utils.py` without modifications

xformers/components/attention/_sputnik_sparse.py
```diff
@@ -185,272 +20,102 @@ def __init__(self, matrix, device=None):
...
-    # for now, our kernels assume that we have the number of
-    # nnz to be divisible by 4
-    mask = _round_nnz(mask[0], divisible_by=4)
-    mask = mask[None].expand(matrix.shape)
-
-    values = matrix[mask].reshape(matrix.shape[0], -1).to(device)
-    row_indices, row_offsets, column_indices = _nonzero_mask_to_sparse_csr_indices(
-        mask[0], device
-    )
-    return values, row_indices, row_offsets, column_indices
```
fmassa, line 456: This has been moved to `xformers/sparse/utils.py` without modifications

xformers/ops.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+import torch
+
+
+def masked_matmul(a, b, mask=None):
+    if torch.overrides.has_torch_function((a, b, mask)):
+        return torch.overrides.handle_torch_function(
+            masked_matmul, (a, b, mask), a, b, mask
+        )
```
fmassa, line 14: This is the key functionality needed for `__torch_function__` to be able to dispatch to custom classes that implement `__torch_function__`

xformers/sparse/csr_tensor.py
```diff
@@ -0,0 +1,437 @@
...
+        v0, v1 = arg0, arg1
+        if isinstance(arg0, cls):
+            v0 = arg0.to_dense()
+        if isinstance(arg1, cls):
+            v1 = arg1.to_dense()
+        out = func(v0, v1)
+        return cls.from_dense(out)
+
+    @classmethod
+    def __torch_function__(cls, func, types, args=(), kwargs=None):
```
fmassa, line 308: All the magic of dispatching lives in here

xformers/sparse/csr_tensor.py
```diff
@@ -0,0 +1,437 @@
...
+            torch.Tensor.__mul__,
+        ]:
+            assert len(args) == 2
+            return cls._binary_op(func, args[0], args[1])
+
+        if func in [torch.Tensor.logical_and, torch.logical_and, torch.Tensor.__and__]:
+            assert len(args) == 2
+            return cls._binary_op_slow(func, args[0], args[1])
+
+        if func in [torch.nn.functional.dropout, torch.dropout, torch.dropout_]:
```
fmassa, line 356: We override `dropout` as well so that everything can be run end to end

xformers/sparse/csr_tensor.py
```diff
@@ -0,0 +1,437 @@
...
+            return cls._wrap(
+                x.shape,
+                x.__values.__deepcopy__(memo),
+                x.__row_indices.__deepcopy__(memo),
+                x.__row_offsets.__deepcopy__(memo),
+                x.__column_indices.__deepcopy__(memo),
+                tuple(v.__deepcopy__(memo) for v in x.__transp_info),
+            )
+
+        if func in [torch.Tensor.grad.__get__, torch.Tensor._grad.__get__]:
```
fmassa, line 410: So that `.grad` returns a `SparseCSRTensor`

xformers/sparse/csr_tensor.py
```diff
@@ -0,0 +1,437 @@
...
+            return cls._wrap(
+                x.shape,
+                x.__values.detach(),
+                x.__row_indices,
+                x.__row_offsets,
+                x.__column_indices,
+                x.__transp_info,
+            )
+
+        if func == torch.Tensor.__deepcopy__:
```
fmassa, line 398: Needed for `nn.Module.state_dict`.

tests/test_timm_wrap.py
blefaudeux: still required given the timm dependency fix ? (pulling timm from main)

xformers/sparse/csr_tensor.py
```diff
@@ -0,0 +1,437 @@
...
+    _diffsort,
+    _get_transpose_info,
+    _transpose_with_info,
+)
+
+
+class SparseCSRTensor(torch.Tensor):
+    @staticmethod
+    def __new__(cls, row_offsets, column_indices, values, shape):
+        kwargs = {}
```
blefaudeux, line 23: nit: any benefit in post populating the dict vs. instantiating it directly ? (kwargs = {"key" : val})

blefaudeux:
LGTM, thanks @fmassa ! It's nice that the dispatch is in a single place, but it feels like it will take some mental space if generalized. On the user side it's certainly awesome that it's transparent and clean, probably worth it.
Could you have a look at the unit test coverage regression ? Looks like it's in [sputnik](https://app.codecov.io/gh/facebookresearch/xformers/compare/124/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=facebookresearch#diff-eGZvcm1lcnMvY29tcG9uZW50cy9hdHRlbnRpb24vX3NwdXRuaWtfc3BhcnNlLnB5)

tests/test_timm_wrap.py
fmassa: I've added `timm` back in the `requirements` file just to continue testing it as before (and be the least BC-breaking as possible). But this now works even if we remove it, which was the original goal when I did this.

I can revert those changes if you want, but I still think keeping test dependencies small is a plus. Up to you

fmassa:
> Could you have a look at the unit test coverage regression ? Looks like it's in sputnik_sparse

I looked into it and it seems to be from before my changes (when we changed the normalization to happen at the query level and not at the attention level). Given the 1.8% regression in coverage that is in there (and given that I plan on removing that whole file sometime in the future), I would consider this as ok.

But happy to add extra tests if you think this is necessary.

xformers/sparse/csr_tensor.py
blefaudeux: Completely, I prefaced it with "nit" because of that, I was wondering whether there was an underlying reason that I did not know

blefaudeux:
I think that it would be nice to improve test coverage in general in the future, but not blocking right now IMO

blefaudeux (approved):
LGTM, thanks @fmassa !

PR #144, opened 2021-12-09
[feat] Dropout(Activation(x+bias)), now with partial BW fusion

blefaudeux:
## What does this PR do?
This was a long time in the making.. Fusing the BW part of the activation/bias/dropout kernel. Not quite perfect but in some places the speed goes really bananas (like x3 or x4 the naive calls).
Fusing this implied flipping the whole problem upside down, basically the seeds have to be per collum, and the kernels (fw and bw) also work that way. This allows us to fuse the bias gradient computations, since it's a sum over that direction

TODO:
- [x] add more unit tests to check that the dropout drops are respected on average
- [x] possibly make sure that the rand mask does not repeat (may or may not be a big deal). Ok this is doable by making the kernels cooperate on the same col, like Phil does on LayerNorm
- [x] improve on the scheduling for small buffers
- [x] Fix the atomic add funkiness (works for now but this does not look completely right, num_warps dependent)
 

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


blefaudeux:
Dropout_Bias_False_FW+BW_torch float16_Act:_gelu
Dropout_Bias_True_FW+BW_torch float32_Act:_gelu
67650d506.png)

Interested @suchenzang ? This took a while to get right. Some speed for small tensors should be recoverable, I didn´t play with the settings too much

suchenzang:
Oh man, coming to xFormers to shop for parts is great.  @blefaudeux these numbers are on V100s or A100s?

blefaudeux:
> Oh man, coming to xFormers to shop for parts is great. @blefaudeux these numbers are on V100s or A100s?

ahah, I missed you @suchenzang ! This is on an ampere laptop, working with what I have around.. 400GB/s is the max bandwidth, so basically there's not much to win on the inference side. The reported training number is not exact GB wise (there are operations in the middle not counted), but could well be that the scheduling brings back another 10/20%, these are very raw numbers. It should be at parity with pytorch accuracy wise, not compromising here, just fusing the kernels. I'll pull in @Chillee also in that he has a NVFuser solution for the same part that you could test out (we were wondering whether xformers could host that also), he was actually the incentive for me to revisit this (he's got very good numbers !)

blefaudeux:
Dropout_Bias_True_FW_torch float32_Act:_None
the scheduling probably has some margin for improvement because fp32 saturates the bandwidth better. Anyway, that's the easy part

blefaudeux:
cc @min-xu-ai , if it helped to have a look at a Triton kernel for Mevo

blefaudeux:
I'll clean up the PR, sorry for all the extra changes

blefaudeux:
> I'll clean up the PR, sorry for all the extra changes

I just pushed a cleaned up version, should be better

xformers/triton/dropout.py
```diff
@@ -25,42 +25,48 @@
 class _dropout(torch.autograd.Function):
     @staticmethod
     @custom_fwd(cast_inputs=torch.float16)
-    def forward(ctx, x, p, bias, activation, activation_grad):
+    def forward(ctx, x, p, bias, activation, activation_grad, trainable_bias):
```
blefaudeux, line 28: the trainable bias (or not) was not properly handled before (bias was always assumed to be trainable, which is mostly true but not always)

ptillet:
@blefaudeux Amazing  ! :raised_hands:

suchenzang:
> > Oh man, coming to xFormers to shop for parts is great. @blefaudeux these numbers are on V100s or A100s?
> 
> ahah, I missed you @suchenzang ! This is on an ampere laptop, working with what I have around.. 400GB/s is the max bandwidth, so basically there's not much to win on the inference side. The reported training number is not exact GB wise (there are operations in the middle not counted), but could well be that the scheduling brings back another 10/20%, these are very raw numbers. It should be at parity with pytorch accuracy wise, not compromising here, just fusing the kernels. I'll pull in @Chillee also in that he has a NVFuser solution for the same part that you could test out (we were wondering whether xformers could host that also), he was actually the incentive for me to revisit this (he's got very good numbers !)

Re: nvfuser - some of our numerics / training dynamics changed with

Chillee:
@suchenzang Out of curiosity, is it easily to isolate the problematic function you used NVFuser with that changed training dynamics?

Also, this is an example of how to use it: https://colab.research.google.com/gist/Chillee/c7667dd02fa08fb62edf1b77285f0374#scrollTo=G7CGthC7cLJo

I'll look into integrating it with xformers sometime soonish.

suchenzang:
@Chillee My bad, jumped the gun on posting too quickly on that, which is why I ended up deleting the comment.  @myleott ended up getting to bottom of this and it involves commenting out `torch._C._jit_set_profiling_executor(False)` in https://github.com/NVIDIA/Megatron-LM/commit/dba2506d11f1250199310993ef694e834442b883 in order to preserve the same training dynamics between v2.4 and v2.6 of Megatron.  Still unclear why this happens, so we will need to circle back on this after our deliverables this half.

blefaudeux:
> Dropout_Bias_False_FW+BW_torch float16_Act:_gelu
> 
> Dropout_Bias_True_FW+BW_torch float16_Act:_squared_relu
> 
> 
> Dropout_Bias_True_FW_torch float16_Act:_squared_relu
> 
> 
> Interested @suchenzang ? This took a while to get right. Some speed for small tensors should be recoverable, I didn´t play with the settings too much
> 

I need to update these, the latest numbers are even better actually, almost all inference is at 350/400GB/s on a laptop

suchenzang:
@blefaudeux let me know when this is ready & merged (& released, ideally!). Will be ready to pull in and test then (no rush).

blefaudeux:
before/after
looks like the small size + training + bias degraded, looking into that. Everything else got a big bump in terms of speed


Screenshot from 2021-12-13 17-28-41
Screenshot from 2021-12-13 17-28-13
Screenshot from 2021-12-13 17-27-29

blefaudeux:
Probably too long on that, it was on the side and I spent a lot of time getting back into the proper context all the time (+couple of small hiccups with triton on my way). Now with good enough perfs I think, there's one case which is not faster than pytorch (really small buffers + gelu + fp16), everything else is significantly faster and the FW speed almost doubled with this PR. I updated all the graphs, keep in mind when comparing that the previous ones were with V100 (HBM memory, something like 900GB/s max bandwidth) and the new ones are with a 3080 laptop (GDDR6 memory, something like 440GB/s max bandwidth). I think that this could be revisited with a newer Triton, or with NVFuser/functorch (if not too much work a PR would be great @Chillee, if the speed is there OOB I would definitely take it !)

Up for review and next is https://github.com/facebookresearch/xformers/issues/153, probably much more impactful if this is doable

blefaudeux:
testing a training with microGPT, looks like sometihng is wrong, the loss plateau..

blefaudeux:
> testing a training with microGPT, looks like sometihng is wrong, the loss plateau.. edit: checked, the stats are wrong on the % of dropout, randint4x is not returning what I thought it was.. fixing that

fixed, no impact on perf. Checking with loss curves right now that everything is fine but should be the case. I've improved on the unit test to catch that, basically p=0.5 was correct but p=0.1 was not

blefaudeux:
Screenshot from 2021-12-18 15-39-03
good, now training properly

suchenzang:
> Screenshot from 2021-12-18 15-39-03
> 
> good, now training properly

Do you have a baseline plot to compare against?

blefaudeux:
> > Screenshot from 2021-12-18 15-39-03
> > good, now training properly
> 
> Do you have a baseline plot to compare against?

this is against the previous main (still FusedMLP), I'm running one right now with MLP instead (pure pytorch). The end test (sampled text) looked good

suchenzang:
> 

So the only mildly worrisome thing here is seeing divergence as training progresses. It seems like that delta shows up a bit vs the blue. If the two converges to different points, then we have a problem :(

blefaudeux:
> So the only mildly worrisome thing here is seeing some divergence as training progresses. It seems like that delta shows up a bit vs the blue. If the two converges to different points, then we have a problem :(
totally, it's very strange because there's no shortcut in the implementation, it's supposed to give the same results. I've tried AMP/fp16, it's not that, same results. Now it could just be about the seed, I'll check that next

blefaudeux:
> > So the only mildly worrisome thing here is seeing some divergence as training progresses. It seems like that delta shows up a bit vs the blue. If the two converges to different points, then we have a problem :(
> 
> totally, it's very strange because there's no shortcut in the implementation, it's supposed to give the same results. I've tried AMP/fp16, it's not that, same results. Now it could just be about the seed, I'll check that next

hmm, testing seeds it does change the result a bit, but still. With this PR we're using less seeds within the random number generation, and generating more numbers out of them, but that's the only measurable difference top of head with the previous take, the rest is just about a different kernel architecture

blefaudeux:
Discussing with Phil, could be that the RNG is not good enough, checking that next

blefaudeux:
> Discussing with Phil, could be that the RNG is not good enough, checking that next

could well be the reason, the fused dropout on main/ is fine. I checked if there was a pattern in the dropout, but not obvious on a heatmap. Dropping that for now, could be that this was the main reason for a very fast FW (generating less seeds), but cannot be at the expense of training accuracy

Issue #157, opened 2021-12-18
pip install broken on OSX and Ubuntu 20.04 (container w/ CUDA) python 3.9


jramapuram:
Not sure if there are any specific requirements or why this particular error occurs, but I see the same error on both:

1. OSX Monterey with a mini forge env -- python 3.9
2. Ubuntu 20.04 in a docker container (with CUDA installed) -- python 3.9

Error is also quite non-descriptive: `assert len(sources) > 0` 

[Full gist](https://gist.github.com/jramapuram/c4a8e6f372a6fceb96d88c09e0b27cad).

dianaml0:
If you [build from source](https://github.com/facebookresearch/xformers#build-from-source-dev-mode) that should fix the issue. This is happening because the files for building extensions aren't there. Will look into improving this.

jramapuram:
Thanks for the quick response @dianaml0. 

- With OSX I get an `clang: error: unsupported option '-fopenmp'`:  I tried installing `llvm`, `gcc` and `libomp` with brew, setting `CC` and `CPP` and none of these worked.  
- With the Ubuntu 20.04 container: the source compile works, but after `pip install -e .` it seems like the cloned repo `xformers` needs to be present (i.e. an `rm -rf xformers` fails to properly import the package: `python -c "import xformers"` --> no ModuleFound). It is quite a strong requirement for the cloned `xformers` repo to be present locally for operation 😬

blefaudeux:
> Thanks for the quick response @dianaml0.
> 
>     * With OSX I get an `clang: error: unsupported option '-fopenmp'`:  I tried installing `llvm`, `gcc` and `libomp` with brew, setting `CC` and `CPP` and none of these worked.
> 

this could be a path issue, MacOS ships its own clang. Have you tried conda by any chance ?

>     * With the Ubuntu 20.04 container: the source compile works, but after `pip install -e .` it seems like the cloned repo `xformers` needs to be present (i.e. an `rm -rf xformers` fails to properly import the package: `python -c "import xformers"` --> no ModuleFound). It is quite a strong requirement for the cloned `xformers` repo to be present locally for operation grimacing

This is the very definition of -e, it means "editable" if I remember correctly, if you install with that and modify the library code then this modified code will be used. `pip install .` or `python3 setup.py install` if you want a fire and forget strategy, top of head (this is not specific to xformers, it's general python/pip principle). We should probably explain this better in the README

jramapuram:
[...]

PR #159, opened 2021-12-20
Improve installation instructions on OSX

fmassa:
## What does this PR do?
Fixes (part of) #157.

We need to specify those flags for installation on OSX to work. This is the same of what we do for torchvision in https://github.com/pytorch/vision#installation

Maybe the `MACOSX_DEPLOYMENT_TARGET=10.9` part can be removed in latest OSes, but further investigation is needed to see if we indeed can remove it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ There are two ways you can install it:
   cd xformers
   pip install -r requirements.txt
   pip install -e .
+  # or, for OSX
+  MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ pip install -e .
   ```
 
 #### Sparse attention kernels


blefaudeux (approved):
Thanks for the quick fix @fmassa !

erdem:
I don't know why PR's OSX compiler instructions were removed from the main branch `README.md` file but I fixed my OSX issue in the same way. 

```
clang: error: unsupported option '-fopenmp'
```

I was getting this error from the compiler. 

However, I solved the issue by setting `CC` and `CXX` with the correct version of `gcc`. Note that, I installed `gcc` with brew.  

`brew list gcc`

**output:**
```
/opt/homebrew/Cellar/gcc/12.2.0/bin/aarch64-apple-darwin21-c++-12
/opt/homebrew/Cellar/gcc/12.2.0/bin/aarch64-apple-darwin21-g++-12
/opt/homebrew/Cellar/gcc/12.2.0/bin/aarch64-apple-darwin21-gcc-12
/opt/homebrew/Cellar/gcc/12.2.0/bin/aarch64-apple-darwin21-gcc-ar-12
...
```
 
**my build command**
```
CC=gcc-12 CXX=g++-12 pip install -e .
``` 
 
 Some helpful comments: 
 https://github.com/ninja-build/ninja/issues/1463#issuecomment-1119514124
 https://github.com/jts/nanopolish/issues/789#issuecomment-641655404

PR #164, opened 2021-12-23
[feat] Dropout partial bw fusion (second take)

blefaudeux:
## What does this PR do?
TL:DR: 
Dropout(activation(x+bias)) layer, now with partially fused BW and much better FW perf (bandwidth bound mostly)
Previously broken in #144 since the end accuracy was not as good as pytorch's. Now fixed

The longer version:
Re-opening #144, I finally got the time to look into that (being blocked on the mem efficient attention by a triton bug :) ). So the issue was in the random number generation, which calls tl.rand4intx from Triton, starting from a random seed that Pytorch generates. This seems to break if the number of generated numbers is too big. This PR introduces a processing per tile, each of the tiles being covered by one seed initially + these random numbers. The solution was just to limit the size of the tiles / use more seeds, and adjust the other hyperparameters accordingly (number of warps), the perf is not really changed.

- Training on minGPT
(orange is the previous take, red is pure pytorch MLP, green is MLP with this layer, scale is log)
Screenshot from 2021-12-22 20-51-50

- one epoch of minGPT:
```
Friends of my soul and flatterers, your forfeit,
Go, give me leave, and with my comfort than
'Tis mine own greatness.

SICINIUS:
I know you not.

MENENIUS:
Consider that which not.
Being once the tackless rock about his soldier,
Is all the fiery ports of this case,
But only in your resolve I have sold to Ravenspurgh,
And not omity my father's death with me?
Wife, and there I will take such on him
That you shall steal me on the business
And then was cruel for it; for in his pleasure,
Beyond the whole of her faults, and full of farewells;
For welcome but thy father return'd could not,
Call their defence and them accuse their wifes,
Suppose two and to be glads.

First Citizen:
He says let Turkey, will not be altered; since led
you on the foundations?

MISTRESS OVERDONE:
Look, what I have done this holy word to fine
He would contemplate; and there my father
Hath the assistantied by as the chair
Where is he was fulfilled: and his way is rather
Hath some his bonnet state, and in his face
```

- some perf curves (if bias is False then the perf gap grows even more)
-- training
Dropout_Bias_True_FW+BW_torch float16_Act:_gelu

Dropout_Bias_True_FW+BW_torch float16_Act:_squared_relu

-- forward pass only (still with dropout)
Dropout_Bias_True_FW_torch float16_Act:_gelu

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


blefaudeux:
cc @ptillet @suchenzang, should finally be ready and perfs are still good. Turns out randint4x from Triton works fine as long as you don´t request too many numbers

blefaudeux:
hmm testing on a desktop 3080, the perf for < 2048 buffers is not very good, and it's not just hyper params. The perf above that is a lot better than before, but would be nice to have the best of both worlds...

xformers/triton/dropout.py
```diff
@@ -19,49 +19,58 @@
...
-        # seed max is int32 max for positive numbers: 2**16
-        seeds = torch.randint(65536, (x_.shape[0],), device=x.device).to(torch.int32)
+        assert bias is None or (bias.dtype == x.dtype and bias.shape[0] == N)
 
-        # SPMD launch grid
         def grid(meta):
             return (
-                x_.shape[0],
-                triton.cdiv(x_.shape[1], meta["BLOCK_SIZE"]),
+                triton.cdiv(M, meta["BLOCK_M"] * 4),
```
dianaml0, line 42: How come the number 4 here?

xformers/triton/k_dropout.py
```diff
@@ -10,138 +10,232 @@
...
+
+            act_grad = meta["ACTIVATION_GRAD"](inputs).to(grad_out.dtype)
+            grad_out *= act_grad
+
+        # randomly prune and scale
+        if p > 0.:
+            # generate all the random numbers for the block at once, then reshape
+            keep = tl.reshape(rand_mask, grad_out.shape)
+
+            # prune and normalize in one go
```
dianaml0, line 217: nit: seems like there's a lot of duplication with the fwd, maybe easy to just pull out into a common method?

dianaml0 (approved):
Looks great! :)

xformers/triton/dropout.py
blefaudeux: Great question, I'll add a note ! It's because optimally the rand generator from Triton returns 4 blocks (you can see that in the kernel, we manually go through them since indexing is not possible), so the real footprint here is 4 times this base setting

xformers/triton/k_dropout.py
blefaudeux: I'll clean that up, you're right, I'll need too kernels still but I can factorize a couple of functions

blefaudeux: @dianaml0 I could not change that much in the end, I just gave it a try, the triton jit phase chokes on some of the factorization options

blefaudeux: ok, fixed by changing some of the formulations, and small speedup on the dropmask generation at the same time. This should be covered, running one last full GPT run and landing if that's ok

Issue #41, opened 2021-10-26
[feat] Compositional Attention


blefaudeux:
# 🚀 Feature
Intriguing paper, keep the softmax(QKt) and V untangled, in that retrievals (*V_i in the vanilla attention) can have a look at all the searchs, that is it can be evaluated against all the softmax(QKt)_j, on a per head basis ("heads" become how many searchs and and many retrieval you support, possibly different)

## Motivation
Interesting take for some tasks, does not seem life changing for classical MLM but seems very relevant to reasoning or vision related tasks

## Pitch
Implement this, see how it goes in something like Dino ?

## Alternatives
Not doing it

## Additional context
[Paper](https://arxiv.org/pdf/2110.09419v1.pdf)
[Reference implementation](https://github.com/sarthmit/Compositional-Attention)

PR #178, opened 2022-01-10
[feat] Compositional attention

blefaudeux:
## What does this PR do?
Implements Compositional Attention (based on the reference implementation), as mentioned in https://github.com/facebookresearch/xformers/issues/41

[Paper](https://arxiv.org/pdf/2110.09419v1)

TODOs
- [x] Sane defaults
- [x] Speedup wherever possible. Looks like it takes a lot of memory also at the moment, probably some dummy mistakes
- [x] Maybe self-attention optimization (single proj) -> doable if moving the projections within the attention to the inproj class, worth it? 
- [x] Add a lot of explanations/documentations
- [0] Some IR results ? -> that would be for another task probably ?

cc @sarthmit if interested 

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


sarthmit:
Please feel free to reach out. I'd be happy to help :)

blefaudeux:
> Please feel free to reach out. I'd be happy to help :)

@sarthmit actually, I have a question: I've seen quite a few .contiguous().do_something() ([for instance](https://github.com/sarthmit/Compositional-Attention/blob/main/fairseq/fairseq/modules/compositional_attention.py#L307)), it's not completely obvious to me that it's beneficial since you change the tensor structure right after that, and you don't seem to be using specific kernels which would require having a contiguous tensor in the first place (there are actually a bunch of kernels in xformers which would love that, but .contiguous would be last in that case, not first). 
Is there another reason that I'm missing, TPUs for instance ?

sarthmit:
@blefaudeux I think you are right, there is probably no use for those .contiguous(). I just used the fairseq multi-head codebase as the starting place and they seem to have it also, so I never bothered to remove it. See: [multi-head](https://github.com/pytorch/fairseq/blob/e69f1fa37fa2bb0099763e9e3649ae56c2850a58/fairseq/modules/multihead_attention.py#L244)

xformers/components/multi_head_dispatch.py
```diff
@@ -127,6 +127,7 @@ def forward(
         key: Optional[torch.Tensor] = None,
         value: Optional[torch.Tensor] = None,
         att_mask: Optional[torch.Tensor] = None,
+        key_padding_mask: Optional[torch.Tensor] = None,
```
fmassa, line 130: do we need `key_padding_mask`? Can't this be obtained through the `att_mask` alone?

blefaudeux: I removed it in an upcoming version, I'm guessing that @sarthmit had it because fairseq, and it complicates the code which is already branchy enough I think

blefaudeux:
microGPT test (causal, autoregressive)

> Friends of my soul: set down.
> What stands before I came to come?
> 
> BENVOLIO:
> Madam, her were born.
> 
> MERCUTIO:
> Tut, such more sorrow we did not so long.
> I have no stand of small as a torment for a
> point a man else.
> 
> ANTONIO:
> Where is the brother of him?
> 
> SEBASTIAN:
> Or despite of all the sun.
> 
> ANTONIO:
> Go, marry, of many choice, my mother,
> Would the near I let thee be freed:
> As with, sir, a brave form'd her by my mood
> Shall twenty years a traitor to turn: tell them
> Show'd their misfer when seems but when I shall never
> Their princely hate with the stealing death.
> 
> DUKE OF AUMERLE:
> I do fear thee, my daughter not stay.
> 
> KING RICHARD II:
> Say that's not swine may know thou art.
> 
> QUEEN ELIZABETH:
> To thou art deceived: if thou both part to tell,
> Thou wouldst be thing in she foul born in presence.
> If I know not, Grumio that thought it good
> Is mutually: here's a stale, if thou beholdst him
> Are our bodies cousin with customary here,
> A fine schoolmaster; there be we fined,
> No, nor do as so fast; it is too long

But loss not as good as a simple attention with the defaults (it's not really the target for this attention mechanism though): 
Screenshot from 2022-01-13 21-20-29

BENCHMARKS.md
```diff
@@ -11,7 +11,7 @@ Please note that:
 - These numbers are dependent of hyperparameters (dimensions chosen for Linformer, sparsity of the pattern), they are mostly an illustration
 - The sparse attention patterns tested here are just presets, as explained in the linked notebook generating any new sparse attention pattern should be relatively easy, while keeping the benefits of optimized computations.
 
-Some examples, generated with `python3 xformers/benchmarks/benchmark_encoder.py --activations gelu --plot -emb 256 -bs 32 -heads 16`
+Some examples, generated with `python3 xformers/benchmarks/benchmark_encoder.py --activations gelu --plot -emb 256 -bs 8 -heads 4`
```
blefaudeux, line 14: reducing the memory load, generating a new graph for everyone

xformers/benchmarks/LRA/code/config.json
```diff
@@ -80,7 +80,7 @@
             "eval_frequency": 50,
             "num_train_steps": 10000,
             "num_eval_steps": 62,
-            "gradient_accumulation": 1
+            "gradient_accumulation": 2
```
blefaudeux, line 83: compositional takes more memory, so this task does not pass on a 8GB gpu (desktop 3080) without bumping up the accumulation

xformers/components/__init__.py
```diff
@@ -48,6 +48,11 @@ def build_multi_head_attention(
                     "num_heads"
                 ]
 
+            if "dim_model" not in multi_head_config["attention"]:
```
blefaudeux, line 51: convenience, remove the need for field duplication in between attention and MHA

xformers/components/attention/compositional.py
```diff
@@ -0,0 +1,341 @@
...
+        self.dim_model = dim_model
+        num_rules = _either_or(num_rules, num_heads)
+        dim_selection = _either_or(dim_selection, dim_model // num_heads)
+
+        # All the initial definition plumbing
+        dim_attn = _either_or(dim_attn, dim_model)
+        dim_key = _either_or(dim_key, dim_model)
+        dim_value = _either_or(dim_value, dim_model)
+
+        self.in_proj_container = (
```
blefaudeux, line 130: difference with the original implementation: reuse xformers input projection, because the task is the same, and there are some optimizations to be had here (if self attention, project once to get q,k,v instead of three calls for instance)

xformers/components/attention/compositional.py
```diff
@@ -0,0 +1,341 @@
...
+        )  # [B * num_heads, Sk, dim_head]
+        v = v.view(B, -1, self.num_rules, self.value_dim).movedim(2, 1).flatten(0, 1)
+
+        # Compute the search: Softmax(QKt)
+        attn_weights = torch.bmm(q, k.transpose(1, 2))  # [B * self.num_heads, Sq, Sk]
+
+        if att_mask_additive is not None:
+            attn_weights += att_mask_additive.values
+
+        if _is_triton_available:
```
dianaml0, line 289: Could we just use the softmax [here]( https://github.com/facebookresearch/xformers/blob/dd96b8d8beda5308fb433c1ef3ff04b7f178c263/xformers/components/attention/core.py#L93)?

dianaml0 (approved):
Interesting approach!

sarthmit:
> @sarthmit if you have cycles for a review, that would be great ! I've removed a few options (not that many actually) and aligned the terms with the other attentions here, + reused existing building blocks that we have (input projections for instance, there's an optimization to be used for self-attention

@blefaudeux What do you mean by cycles for a review? 

By the way, there is also an easy way to test/debug the code. If you set the Value Scores in Equation 13 to the Identity matrix ([here](https://arxiv.org/pdf/2110.09419.pdf)), then you should recover the standard Multi-Head Attention performance.

blefaudeux:
> > @sarthmit if you have cycles for a review, that would be great ! I've removed a few options (not that many actually) and aligned the terms with the other attentions here, + reused existing building blocks that we have (input projections for instance, there's an optimization to be used for self-attention
> 
> 
> 
> @blefaudeux What do you mean by cycles for a review? 
> 
> 
> 
> By the way, there is also an easy way to test/debug the code. If you set the Value Scores in Equation 13 to the Identity matrix ([here](https://arxiv.org/pdf/2110.09419.pdf)), then you should recover the standard Multi-Head Attention performance.

I meant "brain cycles" :) (aka free time). I can have a look later today, but the logic should be maintained

xformers/components/attention/compositional.py
blefaudeux: good catch, fixing that !

blefaudeux:
> > @sarthmit if you have cycles for a review, that would be great ! I've removed a few options (not that many actually) and aligned the terms with the other attentions here, + reused existing building blocks that we have (input projections for instance, there's an optimization to be used for self-attention
> 
> @blefaudeux What do you mean by cycles for a review?
> 
> By the way, there is also an easy way to test/debug the code. If you set the Value Scores in Equation 13 to the Identity matrix ([here](https://arxiv.org/pdf/2110.09419.pdf)), then you should recover the standard Multi-Head Attention performance.

it's a good idea I think, testing that right now, I'm still seeing a difference in perf, investigating. I can see that there are other differences with the Vaswani paper, for instance the normalization is not the same (/sqrt(dim_head) vs /sqrt(dim model)), but I don't think that's enough to explain the difference

blefaudeux:
> 

ok, confirmed @sarthmit @dianaml0 when forcing the values  score to be identity the perf is in line with MHA: the difference in the above is due to.. the rotary embeddings ! These are not part of this PR but they were used in the comparison above for the classical mechanism, and they explained the gap (without them the two mechanisms are aligned if the value score is identity, as expected). All good for landing I think, and it means that supporting rotary embeddings in compositional attention could be a follow up

sarthmit:
@blefaudeux Awesome! I can try sparing some time if needed, is there anything in particular you want me to look at?

I would also be curious to see if you see performance gains on any of the tasks (without the identity matrix)

blefaudeux:
> @blefaudeux Awesome! I can try sparing some time if needed, is there anything in particular you want me to look at?
> 
> I would also be curious to see if you see performance gains on any of the tasks (without the identity matrix)

I've not checked LRA but that could be interesting to get the scores, even if I think that LRA could do with more and better tasks (see this paper from @xwhan for instance https://arxiv.org/pdf/2112.07210.pdf). Rotary embeddings also seem to make a big difference with NLP, would be nice to support them with compositional, should not be a big change. 

There's a definite perf impact though (at least with our implementations, even when pulling in some semi-optimized parts from xformers), it could be another axis to look at because I think that a lot of folks will "just" bump up the number of heads with the scaled dot product instead if compositional has too much of a speed/memory impact

sarthmit:
> > @blefaudeux Awesome! I can try sparing some time if needed, is there anything in particular you want me to look at?
> > I would also be curious to see if you see performance gains on any of the tasks (without the identity matrix)
> 
> I've not checked LRA but that could be interesting to get the scores, even if I think that LRA could do with more and better tasks (see this paper from @xwhan for instance https://arxiv.org/pdf/2112.07210.pdf). Rotary embeddings also seem to make a big difference with NLP, would be nice to support them with compositional, should not be a big change.
> 
> There's a definite perf impact though (at least with our implementations, even when pulling in some semi-optimized parts from xformers), it could be another axis to look at because I think that a lot of folks will "just" bump up the number of heads with the scaled dot product instead if compositional has too much of a speed/memory impact

Thanks, I will check the paper out.

How big is the perf impact? In favour of MHA or in favour of compositional?

blefaudeux:
> > > @blefaudeux Awesome! I can try sparing some time if needed, is there anything in particular you want me to look at?
> > > I would also be curious to see if you see performance gains on any of the tasks (without the identity matrix)
> > 
> > I've not checked LRA but that could be interesting to get the scores, even if I think that LRA could do with more and better tasks (see this paper from @xwhan for instance https://arxiv.org/pdf/2112.07210.pdf). Rotary embeddings also seem to make a big difference with NLP, would be nice to support them with compositional, should not be a big change.
> > 
> > There's a definite perf impact though (at least with our implementations, even when pulling in some semi-optimized parts from xformers), it could be another axis to look at because I think that a lot of folks will "just" bump up the number of heads with the scaled dot product instead if compositional has too much of a speed/memory impact
> 
> Thanks, I will check the paper out.
> 
> How big is the perf impact? In favour of MHA or in favour of compositional?

By perf I meant compute time impact, sorry that wasn't clear. Compositional (this PR at least) is significantly slower than MHA (from this repo again) for the same number of heads, even if I understand that it captures way more relations. Could be worth a second eye on whether there are low hanging fruits?

PR #180, opened 2022-01-17
[feat] make batch dimension broadcastable in MHA

### LINKED ISSUE #87: [feat] Make the batch dimension broadcastable
# 🚀 Feature
Support different key/query/value batch sizes, if 1 or the same value

## Motivation
(popped up when working on [salina](https://github.com/facebookresearch/salina)) PyTorch MHA handles key/query/value with 1 to B batch sizes, will broadcast if in need, We don´t do that and hard crash instead, so the users have to align dimensions from the outside

erip:
## What does this PR do?
Fixes #87

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/tests/test_attentions.py
+++ b/tests/test_attentions.py
@@ -15,6 +15,8 @@ from xformers.components.attention import (
     build_attention,
 )
 
+from typing import Tuple
+
 DEVICES = (
     [torch.device("cpu")] if not torch.cuda.is_available() else [torch.device("cuda")]
 )
@@ -251,6 +253,31 @@ def test_different_kq_dimensions(
 
     res = multi_head(query=q, key=k, value=v)
     assert res.shape == torch.Size([BATCH, seq_q, MODEL])
+
+
+@pytest.mark.parametrize("heads", [1, 4])
+@pytest.mark.parametrize("attention_name", ATTENTION_REGISTRY.keys())
+@pytest.mark.parametrize("device", DEVICES)
+@pytest.mark.parametrize("batch_sizes", [(1, BATCH, BATCH), (BATCH, 1, 1)])
+def test_broadcast_batch_dimension(
+    attention_name: str,
+    heads: int,
+    device: torch.device,
+    batch_sizes: Tuple[int, int, int],
+):
+    Q_BATCH, K_BATCH, V_BATCH = batch_sizes
+    multi_head = _get_multihead(attention_name, 0.0, 0.0, False, heads, device)
+
+    if multi_head.attention.requires_same_k_q_dimensions:
+        # pyre-fixme[29]: The library function `pytest.skip` is not supported by Pyre.
+        pytest.skip(f"{attention_name} does not support different k, q dimensions yet.")
+
+    q = torch.rand((Q_BATCH, SEQ, MODEL), device=device)
+    k = torch.rand((K_BATCH, SEQ, MODEL), device=device)
+    v = torch.rand((V_BATCH, SEQ, MODEL), device=device)
+
+    res = multi_head(query=q, key=k, value=v)
+    assert res.shape == torch.Size([BATCH, SEQ, MODEL])
 
 
 @pytest.mark.parametrize("heads", [1, 4])
--- a/tests/test_attentions.py
+++ b/tests/test_attentions.py
@@ -258,7 +258,7 @@ def test_different_kq_dimensions(
 @pytest.mark.parametrize("heads", [1, 4])
 @pytest.mark.parametrize("attention_name", ATTENTION_REGISTRY.keys())
 @pytest.mark.parametrize("device", DEVICES)
-@pytest.mark.parametrize("batch_sizes", [(1, BATCH, BATCH), (BATCH, 1, 1)])
+@pytest.mark.parametrize("batch_sizes", [(1, BATCH, BATCH), (BATCH, 1, 1), (BATCH, 1, BATCH)])
 def test_broadcast_batch_dimension(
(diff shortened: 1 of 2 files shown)


xformers/components/multi_head_dispatch.py
```diff
@@ -143,8 +143,12 @@ def forward(
...
         self._check(key, "key")
 
-        B, S_Q, _ = query.size()  # Batch x Sequence x Embedding (latent)
-        _, S_K, _ = key.size()  # K, Q's sequence length could differ
+        B, S_Q, EMB = (
+            query.size(0),
+            query.size(1),
+            query.size(2),
+        )  # Batch x Sequence x Embedding (latent) x ...
+        S_K = key.size(1)  # K, Q's sequence length could differ
```
blefaudeux, line 151: we could just use "[expand](https://pytorch.org/docs/stable/generated/torch.Tensor.expand.html#torch.Tensor.expand)" here (not repeat, else memory hog..), in case one of the batch size is not matching. This would have to go with a unit test, to ensure coverage :)

erip: Hmm, I think this is a good idea. I guess the logic is not quite obvious to me since we could have any combination of QKV has batch_size=1, others have batch_size != 1 so we'd need to do some weird checks to make it work, right? Something like:

```python
if query.size(0) == 1 and not (key.size(0) == 1 and value.size(0) == 1):
    query = query.expand_as(...)
elif:
    ...
```

This seems pretty messy. Is there a simpler way to "expand as the largest in dim 0"?

erip: Hmm, also, the `expand` call will break if the sequence lengths differ between Q and K, right?

xformers/components/multi_head_dispatch.py
```diff
@@ -143,6 +143,12 @@ def forward(
         self._check(value, "value")
         self._check(key, "key")
 
+        if query.size(0) < key.size(0):
+            query = query.expand_as(key)
+        elif query.size(0) > key.size(0):
+            key = key.expand_as(query)
+            value = value.expand_as(query)
```
erip, line 150: I don't think this is the best way to do this. Currently it will require the `Q_SEQLEN` and `K_SEQLEN` to match, which is strictly not a requirement... Any thoughts, @blefaudeux?

erip: It also assumes that `key.size() == value.size()` which is probably OK because we already had that assumption, I think.

blefaudeux: hmm I would need to check how the expand_as() works, technically for the computations you don't need Q and K to have the same sequence length so we should not change that. how about 
`
batch = max([q.shape[0], k.shape[0], v.shape[0]])
q , k, v = map(lambda x: x.expand(batch, -1, -1), [q, k v])
`
? top of head, this could be wrong but at least it's not long (who cares about correctness ? :D)

erip: Obvious in hindsight. 😄

xformers/components/multi_head_dispatch.py
```diff
@@ -143,11 +143,8 @@ def forward(
         self._check(value, "value")
         self._check(key, "key")
 
-        if query.size(0) < key.size(0):
-            query = query.expand_as(key)
-        elif query.size(0) > key.size(0):
-            key = key.expand_as(query)
-            value = value.expand_as(query)
+        max_batch = max((query.shape[0], key.shape[0], value.shape[0]))
+        query, key, value = map(lambda x: x.expand(max_batch, -1, -1), [query, key, value])
```
erip, line 147: As an opt, this could be (should be?) conditional on whether broadcasting is supported. WDYT?

tests/test_attentions.py
```diff
@@ -253,6 +255,33 @@ def test_different_kq_dimensions(
     assert res.shape == torch.Size([BATCH, seq_q, MODEL])
 
 
+@pytest.mark.parametrize("heads", [1, 4])
```
blefaudeux, line 258: great test, thank you !

xformers/components/multi_head_dispatch.py
blefaudeux: hmm I think it's fine like this, no ? almost a one liner, fixes some usecases without being an eye sore and it does not seem to break anything. I [checked the speed](https://app.circleci.com/pipelines/github/facebookresearch/xformers/914/workflows/6f29b158-d721-4deb-851a-8031a8f4b8d2/jobs/1823/parallel-runs/0/steps/0-111) and it does not break anything on this side either, so fine by me !

Maybe that I misunderstood your question though, but in general simple broadcasting should have worked, it broke here because when we resize QKV to cater for the heads and fold the heads into the batch dimension, we would not respect this batch difference (and your PR fixes that !). This popped up for a RL usecase, where the policy would generate a batch of trajectories but there was a single K to evaluate against (or related, cc @ludc )

blefaudeux (approved):
Looks great to me, thanks @erip ! Waiting a bit so that the other maintainers can have a look, in case I missed something..

dianaml0 (approved):
LGTM!

erip:
cc @jieru-hu @fmassa any thoughts here?

blefaudeux:
> cc @jieru-hu @fmassa any thoughts here?

It's good I guess, I should hav merged earlier, thanks again @erip !

PR #184, opened 2022-01-18
[feat] 1/N Ragged inference - Add tests

nottombrown:
## What does this PR do?

Adds some tests of some primitives that will be useful for making efficient ragged inference triton kernels

### DIFF
--- a/tests/ragged_inference/test_seq_kv_cache.py
+++ b/tests/ragged_inference/test_seq_kv_cache.py
@@ -1,0 +1,310 @@
+import time
+from functools import lru_cache
+from typing import List, Tuple
+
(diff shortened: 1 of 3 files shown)


blefaudeux (approved):
LGTM, thanks @nottombrown ! I think that it's great to start with the tests and get a baseline out of that. @cpuhrsch I would be curious to see whether we can test [nested tensor](https://github.com/pytorch/nestedtensor) there, it also looks very promising.

Tom, I've fixed a trigger or two in the CI, hopefully the next PR should work better

blefaudeux:
cc @fmassa @dianaml0

PR #185, opened 2022-01-18
[feat] 2/N Ragged inference - Add more tests

nottombrown:
We now have an end to end test of the first triton op that we'd like to write: `calculate_scores_via_qk_dotprod`


blefaudeux:
hmm ok, circleci still off, not much more I can do unfortunately. Having at look at the actual PR @nottombrown

blefaudeux (approved):
LGTM, thanks @nottombrown ! Looking forward to the next step

PR #189, opened 2022-01-21
5/N Ragged Inference - Move triton_v2 ragged inference code to new `experimental` directory

nottombrown:
Since this code using `triton_v2` it's currently incompatible with our CI pipeline. This PR moves it to a separate package that can avoid breaking CI while still letting imports work correctly.

Once triton v2 is stable then we can upgrade `xformers` core to 2.0 and pull the `experimental` code into the core package


blefaudeux:
thanks for the PR @nottombrown, having a look asap ! @fmassa @dianaml0  I'll try to put up a PR this week end to move the mem efficient attention to triton 2 and move it to /experimental also, and same could be done with the favor specific kernel which was breaking triton 1.0

nottombrown:
Great! I'll make a separate branch for further changes so as not to collide with this one!

experimental/ragged_inference_v2/ragged_inference_v2/garbage_pad_ragged_acts.py
```diff
@@ -0,0 +1,253 @@
...
+from typing import List
+
+import numpy as np
+import torch
+import triton
+import triton.language as tl
+
+
+@triton.jit
+def add_kernel(
```
blefaudeux, line 16: ok to remove this ? I can update the branch, spring cleanup so that it's easier to build upon down the line

blefaudeux:
see https://github.com/nottombrown/xformers/pull/2 for minor changes, I hope that works

blefaudeux:
interesting, on a ampere laptop some tests do not pass, and others segfault. How HW ready is triton v2.0 @ptillet ?

ptillet:
Oh this is strange. What are the nature of the failures? It's possible that some tests are not configured to skip configs that require too much shared memory

blefaudeux:
Not a shared memory size issue, segfault at JIT time in Triton/compile. (No
errors when installing or importing Triton). Only happens for some tests,
not all of them

blefaudeux:
Gcc 9 and 10, cuda 11.5 by the way

experimental/ragged_inference_v2/ragged_inference_v2/garbage_pad_ragged_acts.py
blefaudeux: done

experimental/ragged_inference/triton_v2_ragged_qk_dotprod.py
```diff
@@ -0,0 +1,348 @@
...
+        stride_out_seq=scores_out.stride(0),
+        stride_out_q=scores_out.stride(1),
+        stride_out_k=scores_out.stride(2),
+        total_ctx_q_across_all_seqs=n_ctx_q,
+        total_ctx_k_across_all_seqs=n_ctx_k,
+    )
+    return scores_out.reshape((n_ctx_q, n_ctx_k))
+
+
+def ragged_qk_dotprod(
```
blefaudeux, line 296: curious to get some perf numbers on that one, even if probably early

experimental/tests/test_seq_kv_cache.py
```diff
@@ -0,0 +1,313 @@
...
+    assert n_key_ctx_per_seq > n_active_query_ctx_per_seq * 10, (
+        "n_active_query_ctx_per_seq must be much larger than "
+        "n_key_ctx_per_seq for our simulator to be useful because "
+        "we round the HBM memory bandwidth for the active_queries and "
+        "for the scores down to zero"
+    )
+
+    bytes_in_keys_per_seq = n_key_ctx_per_seq * d_model_per_gpu * 2  # 2 from bf16
+    bytes_in_keys_total = bytes_in_keys_per_seq * n_seqs
+    hbm_bw_bytes_per_gpu = 1555e9  # 1.5TB/s
```
blefaudeux, line 262: not for this PR, but I would not do this (compare a number to a theoretical one), (a) it's HW specific -how does this test relate to another accelerator ?- and (b) does some suppositions on what's going on -here the data format for instance-.

For other benchmarks we calcul the user-facing throughput, it's also for instance what Phil does [here](https://github.com/openai/triton/blob/master/python/bench/bench_cross_entropy.py#L24), in that you consider the implementation as a black box, and you count the bytes going in and out (at best for instance you read the seqs and write the attention matrix, the rest is history). It's mostly what happens in this code already, but I would 
- count the BW with num elem * elem_size() (and not suppose bfloat16, would be nice to compare across types actually, it can give an idea on how the kernels are compute or bandwidth bound, at least it helped me on other tasks)

- test for a bunch of sizes, from experience there are a lot of possible holes / scheduling, and testing with one size only is like russian roulette ([see for instance](https://facebookresearch.github.io/xformers/tutorials/triton.html), there are [helpers](https://github.com/facebookresearch/xformers/blob/main/xformers/benchmarks/utils.py#L28) for this in the repo if that helps)

I can do that later on if this ends up running locally, brain dump here :)

experimental/ragged_inference/triton_v2_ragged_qk_dotprod.py
```diff
@@ -0,0 +1,348 @@
...
+
+    out_q_block = tl.load(pid_to_out_q_block_ptr + pid)
+    out_k_block = tl.load(pid_to_out_k_block_ptr + pid)
+    out_seq_idx = tl.load(pid_to_out_seq_idx_ptr + pid)
+    in_q_token_offset = tl.load(pid_to_in_q_token_offset_ptr + pid)
+    in_k_token_offset = tl.load(pid_to_in_k_token_offset_ptr + pid)
+
+    # Define indices ranges, we follow the triton convention of prefixing
+    # with "r" to denote a range like "rq" is the range for queries below
+    rq = in_q_token_offset + tl.arange(0, BLOCK_Q)
```
blefaudeux, line 84: nice, "all" (well, a lot of) the magic is there, looks like nothing but super well done I think

blefaudeux (approved):
looks very good to me, I especially like the wrap for the ragged attentions and how they play well with the kernel. Great comments also, will be nice for newcomers ! Thanks a bunch @nottombrown

experimental/ragged_inference/garbage_pad_ragged_acts.py
```diff
@@ -0,0 +1,191 @@
...
+    ctx_idx = tl.program_id(axis=1)
+
+    # This program will process inputs that are offset from the initial data.
+    # for instance, if you had a vector of length 256 and block_size of 64, the programs
+    # would each access the elements [0:64, 64:128, 128:192, 192:256].
+
+    ragged_acts_offset_ptr = ragged_acts_offset_per_seq_ptr + seq_idx
+    ragged_acts_offset = tl.load(ragged_acts_offset_ptr)
+
+    # Create a mask to guard memory operations against out-of-bounds accesses
```
suchenzang, line 37: Is this unavoidable? How much more memory does this consume (if any)?

experimental/ragged_inference/garbage_pad_ragged_acts.py
```diff
@@ -0,0 +1,191 @@
...
+        ragged_acts = self.raw_tensor
+        d_model = ragged_acts.shape[-1]
+        padded_acts = torch.empty(
+            n_seqs, n_ctx_max, d_model, dtype=ragged_acts.dtype, device="cuda"
+        )
+
+        # We just use one program per n_ctx position for simplicity
+        assert d_model >= 128, f"bad {d_model=}"
+        assert d_model <= 8 * 1024, f"bad {d_model=}"
+        assert d_model % 32 == 0, f"bad {d_model=}"
```
suchenzang, line 131: Why 32 and not 64? Where did these requirements come from?

experimental/ragged_inference/seq_kv_cache.py
```diff
@@ -0,0 +1,170 @@
...
+    We cache this because it requires some substantial CPU work and it's done multiple
+    times sequentially (once per resblock)
+    """
+    indices_list = []
+    ragged_idx = 0
+    max_n_ctx = max(n_ctx_per_kv_cache)
+    for n_ctx in n_ctx_per_kv_cache:
+        for idx_into_seq in range(max_n_ctx):
+            if idx_into_seq < n_ctx:
+                indices_list.append(ragged_idx)
```
suchenzang, line 148: Wondering why we need O(n) append calls here...

experimental/ragged_inference/triton_v2_matmul.py
```diff
@@ -0,0 +1,196 @@
...
+                        )
+                    )
+
+    return configs
+
+
+def get_all_configs():
+    return [
+        # basic configs for compute-bound matmuls
+        triton.Config(
```
suchenzang, line 46: Where do these magic numbers / configs come from?

experimental/ragged_inference/triton_v2_ragged_qk_dotprod.py
```diff
@@ -0,0 +1,356 @@
...
+        )
+        k_tile = tl.load(
+            k_ptr_tile,
+            mask=(rd[:, None] < d_max_offset) & k_ctx_in_bounds[None, :],
+            other=0.0,
+        )
+
+        # In einsum notation, the tl.dot does: qd,dk->qk
+        # This should use tensorcores, so the inputs might be fp16, but the outputs
+        # and all the internal accumulators are fp32
```
suchenzang, line 125: Do internal accumulators default to fp32 or tf32 on A100s?

suchenzang:
Just asking a bunch of questions for my own learnings - feel free to ignore!

experimental/ragged_inference/garbage_pad_ragged_acts.py
blefaudeux: it's part of the Triton API (masked load), it should not take any registers really. This is all in SoC space, so nothing visible in the RAM, worst case it's a throwaway mask on the actual gpu chip

experimental/ragged_inference/seq_kv_cache.py
blefaudeux: please submit a PR :D yes this is not optimal indeed

experimental/ragged_inference/garbage_pad_ragged_acts.py
blefaudeux: I think it comes from BLOCK_K = 32, a scheduling/tiling constraint for the matmuls. This could be relaxed if we were to mask over that dimension in the kernel

experimental/ragged_inference/triton_v2_matmul.py
blefaudeux: comes from here, empirical good values for Ampere GPUs, architecture dependent but Triton does navigate around some specifics thanks for all these scheduling options https://github.com/openai/triton/blob/v2.0/python/triton/ops/matmul.py#L35

blefaudeux:
> Just asking a bunch of questions for my own learnings - feel free to ignore!

no worries, very good questions I think, I tried to give some insights but @nottombrown could probably add a little. Note that the code changed a tiny bit since this PR, and some updates were planned

Issue #15, opened 2021-10-20
Sputnik in Xformers


tgale96:
# ❓ Questions and Help

Hello! I'm the author of Sputnik and noticed that you've forked it! I'm interested to understand your use cases. If you have any questions about the library I'm happy to help!

tgale96:
No worries! I'm just happy to see more users :) Fine-grained sparse attention topologies is a cool use case! I'd be really curious to see what kinds of results you're able to get.

For supporting batched CSR where each matrix has its own topology I would do the following. To store the matrix I'd pack all the nonzeros/indices/offsets into the same allocations. Then I'd add another offset array to the data structure to encode the starts of each of the individual matrices. In the kernel I'd do two modifications. First, you can launch a [3D grid](https://github.com/google-research/sputnik/blob/master/sputnik/spmm/cuda_spmm.cu.cc#L582) of thread blocks, where the third dimension encodes which matrix in the batch is to be processed. Then you can wrap the [kernel device function](https://github.com/tgale96/sputnik/blob/main/sputnik/spmm/cuda_spmm.cu.cc#L106) with a small function that will a) lookup the threadblock's matrix in the batch and b) offset the pointers to the start of the sparse/dense matrices and c) call the kernel device function.

FP16 is a little tricky with Sputnik because I never got around to removing the constraint that the number of nonzeros in each row must be divisible by 2. I had another request recently to get rid of this so it seems like maybe I should finally allocate some time for it :)

That's a good point about the index limitation - I'd been primarily using them for models where the weights are sparse and it's less common to have 32k weights in one dimension. I know cuSPARSE supports 32-bit indices with fp16 data but you obviously wouldn't be able to modify these kernels. I'll have to think about what it will take to add support for 32-bit indices with 16-bit data.

Hope this helps! Looks like you've done some great work. I'm excited to look around the repo more!

Trevor

blefaudeux:
closing that for now since it's not actively being worked on, feel free to reopen anytime !

siddharth9820:
Hi @tgale96!
Did you get around to fixing the indexing issue for fp-16?

PR #192, opened 2022-01-24
[feat]Adding a simple blockwise attention

xwhan:
## What does this PR do?
This attention is used in fairseq to build a long-doc summarization model. 

More efficient implementations should be possible with Triton or blocksparse. However, it needs to be compatible with CPU inference for the sake of deployment and demo.

## Before submitting

- [x] Did you have fun?
  - [x] N/A
  - [x] N/A
  - [x] N/A
  - [x] N/A

## PR review
@dianaml0


blefaudeux:
hey @xwhan thanks for the PR ! FYI sparse attention (just pass a block mask) would work also (it's CPU compatible, although not very optimized as of now, cc @fmassa),  curious to see a benchmark in between the two

blefaudeux:
on another front I've just added some more explanations on pre-commit [here](https://github.com/facebookresearch/xformers/blob/main/CONTRIBUTING.md#coding-style), in case that helps ? sorry about all these steps, we're trying to keep the repo coherent and easy to read

xwhan:
> hey @xwhan thanks for the PR ! FYI sparse attention (just pass a block mask) would work also (it's CPU compatible, although not very optimized as of now, cc @fmassa), curious to see a benchmark in between the two. There are pattern preset here if that helps https://github.com/facebookresearch/xformers/blob/main/xformers/components/attention/attention_patterns.py

Thanks, I did not know those are already CPU compatible. I'll try it out with block masks and see if it's faster and stable in my experiments.

xwhan:
@blefaudeux I spent some time trying out passing a blockwise mask to scaled_dot_product_attention. 

```
        repeat = seq_len // self.block_size
        block_mask = torch.ones(self.block_size, self.block_size)
        mask = torch.kron(torch.eye(repeat), block_mask).bool()
```

The fp16 support is the main pain point. It uses 1/4 more memory and is 2x slower than the PR with fp16. When compared them in fp32 settings, the sputnik one saves a little memory but is still much slower. IMO, this fp16 bottleneck of sputnik might block most NLP use cases, especially for large-scale pretraining.

blefaudeux:
> @blefaudeux I spent some time trying out passing a blockwise mask to scaled_dot_product_attention.
> 
> ```
>         repeat = seq_len // self.block_size
>         block_mask = torch.ones(self.block_size, self.block_size)
>         mask = torch.kron(torch.eye(repeat), block_mask).bool()
> ```
> 
> The fp16 support is the main pain point. It uses 1/4 more memory and is 2x slower than the PR with fp16. When compared them in fp32 settings, the sputnik one saves a little memory but is still much slower. IMO, this fp16 bottleneck of sputnik might block most NLP use cases, especially for large-scale pretraining.

fair enough, thanks for checking it out ! fp16 support is a known issue for sparse (see https://github.com/facebookresearch/xformers/issues/15), we don't have that many cycles or manpower to work on that unfortunately. 

Blocksparse handles it though, but in turn it's not CPU compatible (give it a try, it should be pretty fast ! [Some example here](https://github.com/facebookresearch/xformers/blob/main/HOWTO.md#blocksparseattention), [this](https://github.com/facebookresearch/xformers/blob/main/xformers/components/attention/attention_patterns.py#L168) can also be a helper), so IMO your PR has value in this in-between space. Ideally we should have stronger primitives with sparse and blocksparse though, and @fmassa is working towards a single interface for all these

fmassa:
As @blefaudeux mentioned, I'm working on refactoring the sparse abstractions so that they are unified.
I can add a naive CPU implementation that will make it possible to run things on both CPU and well as GPU (through Triton). A WIP branch is in https://github.com/facebookresearch/xformers/tree/blocksparse_refactoring_v2

I will look into it today

fmassa:
Hi @xwhan 

With https://github.com/facebookresearch/xformers/pull/202 merged, we now have native support for blocksparse tensors with both CPU and CUDA support.

fmassa:
Hi @xwhan 

Let me know if you would like some help getting the new `BlockSparse` Tensor integrated in your pipeline. I believe it might be preferable over implementing multiple (similar) variants using the same base code, as all changes needed would just be on defining the sparse matrix

xwhan:
Thanks @fmassa for the update. That would be a great feature. I am closing the PR for now

PR #199, opened 2022-01-28
Optimize sparse transpose

fmassa:
PyTorch now supports stable sort, so we can remove our past workaround that manually implemented stable sort.

On some quick benchmarks, this improves the `_transpose` speed by ~30%.

Tests are already implemented in https://github.com/facebookresearch/xformers/blob/d4b5757fe5a160fb016fb02eb4e74c2dcf75a75b/tests/test_custom_ops.py#L384-L397


dianaml0 (approved):
Nice!

PR #202, opened 2022-02-02
Create BlockSparse Tensor

fmassa:
## What does this PR do?

This PR creates a BlockSparse Tensor (similar to SparseCSRTensor). This will ultimately enable to use the same code-path in the `scaled_dot_product_attention`, so that users just need to pass a block-sparse matrix in the mask to get the expected results (as is the case now with `SparseCSRTensor`). This is not yet entirely plugged, but will be done in a follow-up PR.

I haven't made the `Blocksparse` `nn.Module` use our new Tensor for now as it relies on additional information for the softmax which I would rather not add in the `BlockSparseTensor` API (all the different types of masking).

Note that for now the backward of `masked_matmul` is not passing, which is very weird as it was pretty-much a copy-paste of what we had before. This needs to be further investigated.

PR #208, opened 2022-02-06
Cleanup sparse tests

fmassa:
## What does this PR do?

With #202 merged, we can now have a single entry-point for testing that sparse tensors behave as expected, with a unified set of tests to cover the required operators.

I've removed `test_sparse_csr.py` as it is now redundant compared to `test_sparse_tensors.py`.

I think a next step would be to clean `test_custom_ops.py` so that they test the bare minimum and doesn't depend on the legacy `SparseCS` tensor.


fmassa:
Looks like doc failures are unrelated?

blefaudeux:
> Looks like doc failures are unrelated?

Doc failure now fixed :) unrelated indeed

xformers/sparse/blocksparse_tensor.py
```diff
@@ -114,8 +114,7 @@ def __init__(self, values, layout):
     def __repr__(self):
         return f"block_sparse_tensor(shape={self.shape}, values={self.__values})"
 
-    @property
-    def _blocksparse_values(self):
+    def values(self):
         return self.__values
```
fmassa, line 118: We now unify obtaining the values of the sparse tensor under the same API, which follows the API from PyTorch

fmassa:
Merging, as this is a harmless PR wrt functionality.

PR #210, opened 2022-02-07
Add nd-tensor support for triton softmax

fmassa:
## What does this PR do?

This PR generalizes `xformers.triton.softmax` to support tensors of more than 3 dimensions.
It does it by flattening the input so that it is 3d.

A similar patch was already added in the past so that it supports 2d tensors, so this is follows in a similar way.

I can add more tests if needed


blefaudeux (approved):
LGTM with a one liner addition, please test this case just to make sure we don't break it in the future (and forget about it)

tests/test_triton_softmax.py
```diff
@@ -28,6 +28,8 @@
     (1, 2048, 2048),
     (1, 3136, 3136),
     (1, 4096, 4096),
+    (2, 2, 384, 384),
```
blefaudeux, line 31: thanks !

blefaudeux (approved):
LGTM again, happy to land it anytime

PR #212, opened 2022-02-08
Fix causal flag on softmax Triton on CPU

fmassa:
## What does this PR do?

We had a default dispatch for CPU tensors in the `triton/softmax` implementation. But it didn't handle the `causal` flag properly.
Instead, it raised a warning and returned wrong values.

Instead, let's mask the tensor with a causal mask, so that results match on both CPU and GPU.

### DIFF
--- a/xformers/triton/softmax.py
+++ b/xformers/triton/softmax.py
@@ -19,7 +19,6 @@ from xformers.triton.k_softmax import _softmax, _softmax_backward
 
 
 _triton_registered_overflow = False
-_triton_registered_warnings = False
 _triton_softmax_fp16_enabled = False  # NOTE: PyTorch keeps softmax as fp32
 
 
@@ -179,7 +178,6 @@ def _softmax_dispatch(
     # - there was no previous failure
 
     global _triton_registered_overflow
-    global _triton_registered_warnings
 
     try:
         if torch.cuda.is_available() and x.is_cuda and not _triton_registered_overflow:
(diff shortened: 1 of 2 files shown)


xformers/triton/softmax.py
```diff
@@ -194,16 +192,12 @@ def _softmax_dispatch(
...
-        logging.warning(
-            "Triton softmax could not be used. \
-                The causal flags is being passed but it does not provide any benefit with PyTorch softmax."
-        )
-        _triton_registered_warnings = True
-
     if mask is not None:
         x = x + mask
 
+    if causal:
```
blefaudeux, line 198: it feels like this falls into the "silently support something which probably does not provide benefits", no ? You used to have strong feelings about that.

blefaudeux:
looks ok to me, the wording in the warning was not correct but the behaviour was intended, in that it was "just" not causal if it could not pull the triton backend. The whole point of this is to improve speed though (I don't think that it helps a lot to support causality computations in the attention and in the softmax as such), and pytorch at the moment does not support that. So it looks like one of these "silently slow down" cases that you don't like and typically prefer to assert on, right ?

blefaudeux:
Fine by me, it was just to raise this silent slow fail case, basically I think that the users should know that this is not doing what they might think it is doing, if on cpu. 

Else using this flag does provide a speed benefit with triton, of course, that's the whole point (we have curves for that in the benchmarks)

blefaudeux (approved):
LGTM !

blefaudeux:
> I'm not 100% sure actually, as PyTorch recently introduced a `masked_softmax` [pytorch/pytorch#69268](https://github.com/pytorch/pytorch/pull/69268) (although this is still a private API)

well, I wrote "pytorch at the moment does not support that", so there's no contradiction really ?

fmassa:
> Fine by me, it was just to raise this silent slow fail case, basically I think that the users should know that this is not doing what they might think it is doing, if on cpu.

Hum, I'm not sure I understand your point in here. What is it supposed to be doing? Being faster or doing "causal" computation?

blefaudeux:
> > Fine by me, it was just to raise this silent slow fail case, basically I think that the users should know that this is not doing what they might think it is doing, if on cpu.
> 
> Hum, I'm not sure I understand your point in here. What is it supposed to be doing? Being faster or doing "causal" computation?

that's why I thought that the wording was probably at fault: yes it is faster, if you know that the attention is causal then there's no point in normalizing the upper triangular part of the matrix, so this is  a way (not clear enough !) to reap this benefit (same as the recent pytorch masked softmax)

PR #214, opened 2022-02-08
Make triton optional in BlockSparseTensor

fmassa:
## What does this PR do?

This is a follow-up from https://github.com/facebookresearch/xformers/pull/213

Now that we have fall-back implementations, we don't need to error-out if Triton is not available in the system.

We still raise the warning as before, but at least users can use the CPU / naive CUDA implementations.

### DIFF
--- a/tests/test_sparse_tensors.py
+++ b/tests/test_sparse_tensors.py
@@ -209,7 +209,10 @@ def test_sparse_softmax(tensor_type, device):
     res_sparse.values().backward(torch.ones_like(res_sparse.values()))
     res_gt.backward(torch.ones_like(res_gt))
 
-    assert torch.allclose(a.grad, a_sparse.grad.to_dense(), atol=2e-7)
+    a_grad = a.grad.clone()
+    a_grad[~mask] = 0
+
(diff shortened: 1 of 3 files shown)


tests/test_sparse_tensors.py
```diff
@@ -209,7 +209,10 @@ def test_sparse_softmax(tensor_type, device):
     res_sparse.values().backward(torch.ones_like(res_sparse.values()))
     res_gt.backward(torch.ones_like(res_gt))
 
-    assert torch.allclose(a.grad, a_sparse.grad.to_dense(), atol=2e-7)
+    a_grad = a.grad.clone()
+    a_grad[~mask] = 0
```
fmassa, line 213: This was a "bug" in the test before, as PyTorch implementation of `softmax` yields `inf` / `nan` in the backward if a full row is empty.

This makes me think that we should probably provide a "safe-softmax` alternative in xformers. It won't be very efficient if done with pure pytorch operators, but I've already found myself in need of such a function and thus it would be handy to have it somewhere

blefaudeux (approved):
LGTM, thanks @fmassa !

dianaml0 (approved):
Awesome thanks!

Issue #219, opened 2022-03-01
xformers ViT-B ImageNet training instability compared to timm


jramapuram:
# 🐛 Bug

I'm trying to create a 1:1 config that can train a stable ViT-B with the MAE config (from appendix A.2). 

Maybe I'm missing something (highly plausible), but when I use xformers instead of timm it creates an unstable training scenario [over numerous trials] with exactly the same hyper-parameters (batch_size=4096 + cutmix + mixup + label smoothing + AdamW[0.9, 0.95], lr=1e-4 [with scaling rule ofc], lr warmup + cosine decay, skip bias/CLS/pos_embed weight decay, etc, etc).

![image](https://user-images.githubusercontent.com/8204807/156156416-ca13a631-d5ec-4662-a871-2796e82dddce.png)

## xformers ViT-B Config
```python
reversible: False
block_type: "encoder"
num_layers: 12
dim_model: 768
layer_norm_style: "pre"

multi_head_config:
  num_heads: 12
  residual_dropout: 0.1  # (1) tried without this, (2) swapping this for DropPath, (3) with regular dropout
  use_rotary_embeddings: False

  attention:
    name: "scaled_dot_product"
    dropout: 0.0
    causal: False

feedforward_config:
  name: "MLP"
  dropout: 0.0
  activation: "gelu"
  hidden_layer_multiplier: 4
```

## xformers ViT-B 
```python
"""A simple ViT-B in xformers."""

import typing as t
from pathlib import Path
import yaml

import torch
from torch import nn
from timm.models.vision_transformer import DropPath
from timm.models.layers.patch_embed import PatchEmbed
from timm.models.layers.weight_init import trunc_normal_
from xformers.factory import xFormer, xFormerConfig


def _init_vit_weights(module: nn.Module):
    """Transformer weight initialization from TIMM."""
    if isinstance(module, nn.Linear):
        trunc_normal_(module.weight, std=0.02)
        if module.bias is not None:
            nn.init.zeros_(module.bias)
    elif isinstance(module, (nn.LayerNorm, nn.GroupNorm, nn.BatchNorm2d)):
        nn.init.zeros_(module.bias)
        nn.init.ones_(module.weight)


class ViT(nn.Module):
    """Vision transformer + head module."""

    def __init__(
        self,
        img_size: int = 224,
        in_chans: int = 3,
        patch_size: int = 16,
        num_classes: int = 1000,
        drop_path_rate: float = 0,
[...]

PR #233, opened 2022-03-11
[bug] Pre-norm wrapper only normalizing the first input

blefaudeux:
## What does this PR do?
Could well explain #219 (cc @jramapuram), I stumbled upon this bug by luck while rewriting some of the input projection part and being more strict with self-attention (in that case the tensors need to be the exact same objects, with an incoming PR, makes it easier to catch a bug if the intent is self attention). 

This showed that the pre-norm wrapper was (a) only normalizing the first input (b) creating new objects because it would normalize the tensors one by one, even if it was the same tensor to begin with. This is a bug which already existed in the past and was fixed, not sure how it came back (long lived branch + botched merge maybe), hence the unit test and the change of interface to make sure that this does not happen anymore. Consequence was both correctness + speed (after the pre-norm, the tensors were not the same anymore, so the self attention speed up was off)

(a) I changed the interface to make it compulsory to pass the inputs as a list for all these wrappers, so that there's no more confusion
(b) is unit tested in this PR + the incoming PR will add another test

## Before submitting

- [x] Did you have fun?

## PR review

### DIFF
--- a/xformers/components/residual.py
+++ b/xformers/components/residual.py
@@ -26,8 +26,15 @@ class LayerNormStyle(str, Enum):
     Post = "post"
 
 
+class RequiresWrappedInputs:
+    """Used to mark, through inheritance,
+    the fact that this class will require inputs to be passed as a single list"""
+
+    pass
+
+
 # CREDITS: the following is inspired by FastAI's Transformer implementation
-class Residual(nn.Module):
+class Residual(nn.Module, RequiresWrappedInputs):
     """
     Object-oriented handling of the residual path
 
@@ -49,7 +56,7 @@ class Residual(nn.Module):
             return inputs[0] + self.layer(*inputs, **kwargs)
 
 
-class PreNorm(nn.Module):
+class PreNorm(nn.Module, RequiresWrappedInputs):
     """Adds LayerNorm before computing attention
 
     ..Note: If a list of inputs is passed, all of them get normalized"""
@@ -62,9 +69,7 @@ class PreNorm(nn.Module):
             self.norm = nn.LayerNorm(d_model)
 
         self.sublayer = sublayer
-        self.wrap_inputs = isinstance(sublayer, PostNorm) or isinstance(
-            sublayer, Residual
-        )
+        self.wrap_inputs = isinstance(sublayer, RequiresWrappedInputs)
 
     def forward(self, inputs: List[torch.Tensor], **kwargs):
         assert len(inputs) > 0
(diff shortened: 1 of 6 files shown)


blefaudeux:
As usual, reversible is the pain point.. fixing that

xformers/components/residual.py
```diff
@@ -34,21 +26,37 @@ class LayerNormStyle(str, Enum):
     Post = "post"
 
 
+class RequiresWrappedInputs:
```
blefaudeux, line 29: classes which derive from this class only accept a single input list (makes it impossible to subtly footgun)

xformers/components/residual.py
```diff
@@ -16,14 +16,6 @@
     from xformers.triton.layer_norm import FusedLayerNorm
 
 
-def _to_tensor_list(
```
blefaudeux, line 19: this was supposed to be a helper, but in the end it was masking bugs (in that layer(x, y, z) could have a different behaviour depending on the residual wraps). I think that it's better to force inputs in a single fashion

xformers/components/residual.py
```diff
@@ -34,21 +26,37 @@ class LayerNormStyle(str, Enum):
...
         self.layer = layer
 
-    def forward(self, inputs: Union[torch.Tensor, List[torch.Tensor]], *args, **kwargs):
-        inputs = _to_tensor_list(inputs)
+        # PreNorm and PostNorm require all the tensors to be passed as a list
+        self.wrap_inputs = isinstance(layer, PreNorm) or isinstance(layer, PostNorm)
 
-        return inputs[0] + self.layer(*inputs, *args, **kwargs)
+    def forward(self, inputs: List[torch.Tensor], **kwargs):
+        if self.wrap_inputs:
```
blefaudeux, line 52: the trick here is that these residual/norm wrapper can wrap themselves at times. When they wrap an external layer, then the inputs are unrolled, when the sublayer is another wrap then we maintain inputs=List[Tensor] to prevent bugs like this one

xformers/factory/block_factory.py
```diff
@@ -335,8 +335,8 @@ def forward(
             q, k, v = x, x, x
 
         # Pre/Post norms and residual paths are already handled
-        x = self.wrap_att(q, k, v, att_mask=att_mask)
-        x = self.wrap_ff(x)
+        x = self.wrap_att(inputs=[q, k, v], att_mask=att_mask)
```
blefaudeux, line 338: all the wraps require a single input list + kwargs, which I believe is more future proof (this bug cannot happen, or at least not as easily)

dianaml0 (approved):
Nice catch and nice fix!

jramapuram:
image
@blefaudeux :  Looks within tolerance :)

PR #239, opened 2022-03-17
T112038261 reduce test runtime

coleholcomb:
## What does this PR do?
Decreased CPU test suite runtime in Meta internal testing utility. This was requested via a Meta task filed by @dianaml0.

The changes include:

- decreasing the number of parameterized test cases in some of the highest cardinality test cases
- reducing tensor sizes in some of the slowest running tests (several minute runtime in Sandcastle -> few seconds)
- small number of changes to reduce inefficiencies in the code involving itertools.product()

## Before submitting

- [🙃 ] Did you have fun?
- [X] Did you read the [contributor guideline](https://github.com/facebookresearch/xformers/blob/master/CONTRIBUTING.md)?
- [-] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
  This was requested via a Meta task.
- [-] Did you make sure to update the docs?
  - [X] N/A
- [-] Did you write any new necessary tests?
  - [X] N/A
- [-] Did you update the [changelog](https://github.com/facebookresearch/xformers/blob/master/CHANGELOG.md)? (if needed)
  - [X] N/A

## PR review

### DIFF
--- a/tests/test_attentions.py
+++ b/tests/test_attentions.py
@@ -103,7 +103,7 @@ def test_order_invariance(
     )
 
     # Check that a shuffled input produces the same results
-    seqs = [SEQ, SEQ//2] if (attention_name != "blocksparse") else [SEQ]
+    seqs = [SEQ, SEQ // 2] if (attention_name != "blocksparse") else [SEQ]
 
     for seq in seqs:
         # Check that we can pass a smaller sequence
@@ -194,7 +194,7 @@ def test_inproj(
     if same_settings:
         in_proj = InProjContainer(in_params, None, None)
     else:
-        out_features = MODEL if same_sizes else MODEL//2
+        out_features = MODEL if same_sizes else MODEL // 2
         in_params_flip = InProjParams(MODEL, out_features, not proj_bias, small_init)
         in_proj = InProjContainer(in_params, in_params_flip, in_params_flip)
 
@@ -248,7 +248,7 @@ def test_different_kq_dimensions(
         # pyre-fixme[29]: The library function `pytest.skip` is not supported by Pyre.
         pytest.skip(f"{attention_name} does not support different k, q dimensions yet.")
 
-    seq_q = SEQ//2
+    seq_q = SEQ // 2
     q = torch.rand((BATCH, seq_q, MODEL), device=device)
     k = torch.rand((BATCH, SEQ, MODEL), device=device)
     v = torch.rand((BATCH, SEQ, MODEL), device=device)
--- a/tests/test_attentions.py
+++ b/tests/test_attentions.py
@@ -23,7 +23,7 @@ DEVICES = (
 
 BATCH = 2
 SEQ = 128 if torch.cuda.is_available() else 32
-MODEL = 128 if torch.cuda.is_available() else 64
+MODEL = 128 if torch.cuda.is_available() else 16
 GLOBAL_ATTENTION_RATIO = (
     _DENSITY_THRESHOLD * 0.9
 )  # Make sure that we test the sparse implementation, no matter the threshold
@@ -81,7 +81,7 @@ def _get_multihead(
     return multi_head
 
 
-@pytest.mark.parametrize("attn_dropout", [0.0, 0.3])
+@pytest.mark.parametrize("attn_dropout", [0.3])
 @pytest.mark.parametrize("residual_dropout", [0.0, 0.1])
 @pytest.mark.parametrize("causal", [True, False])
 @pytest.mark.parametrize("heads", [1, 4])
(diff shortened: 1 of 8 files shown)


tests/test_attentions.py
```diff
@@ -81,7 +81,7 @@ def noop(x):
     return multi_head
 
 
-@pytest.mark.parametrize("attn_dropout", [0.0, 0.3])
```
dianaml0, line 84: I think it would still be valuable here to test the case of no dropout

tests/test_nystrom_attention.py
```diff
@@ -13,14 +13,14 @@
 
 @pytest.mark.parametrize("pinverse_original_init", [True, False])
 @pytest.mark.parametrize("use_razavi_pinverse", [True, False])
-@pytest.mark.parametrize("num_landmarks", [30, 33, 905])
```
dianaml0, line 16: I think it's still valuable to test this because we test different code paths, one where the sequence length is divisible by the number of landmarks and one where it isn't

tests/test_attentions.py
```diff
@@ -81,7 +81,7 @@ def noop(x):
...
+@pytest.mark.parametrize("attn_dropout", [0.3])
```
blefaudeux, line 84: 0 and 0.3 are different codepaths (0 will skip dropout completely for most implementations), I think we should keep both for safety reasons

tests/test_attentions.py
```diff
@@ -103,7 +103,7 @@ def test_order_invariance(
     )
 
     # Check that a shuffled input produces the same results
-    seqs = [SEQ, SEQ - 16] if (attention_name != "blocksparse") else [SEQ]
+    seqs = [SEQ, SEQ // 2] if (attention_name != "blocksparse") else [SEQ]
```
blefaudeux, line 106: yop good point, that works !

tests/test_compositional_attention.py
```diff
@@ -29,10 +29,10 @@
 assert ATTENTION_REGISTRY.keys(), "Attention layers should have been registered"
 
 
-@pytest.mark.parametrize("attn_dropout", [0.0, 0.3])
+@pytest.mark.parametrize("attn_dropout", [0.3])
```
blefaudeux, line 32: same as before, we should keep 0 because different codepath down the line ([see](https://github.com/facebookresearch/xformers/blob/main/xformers/triton/dropout.py#L171) for instance)

tests/test_global_attention.py
```diff
@@ -31,7 +31,7 @@ def test_ratio(global_attention_ratio: float):
         # Check that the tokens which have access to the full attention give the same
         # results as the monolithic dense scaled_dot_product
         mask = config["attention_query_mask"][:, 0]
-        torch.allclose(r_global[:, mask, :], r_dense[:, mask, :])
+        assert torch.allclose(r_global[:, mask, :], r_dense[:, mask, :])
```
blefaudeux, line 34: wops, thanks for the catch here

blefaudeux (commented):
LGTM except for the dropout, skipping the dropout call when the prob is 0 is an easy optimization that a lot of implementation use (including our own triton layers), I would put this back. Other than that this is great for me, thanks a bunch @coleholcomb !

tests/test_nystrom_attention.py
coleholcomb: Currently I have [33, 905], would [30, 905] be preferable? This is a somewhat slow test, so I'm trying to reduce the test cases.

coleholcomb: Based on this feedback, I refactored this test into two separate tests. The "close_to_sdp" test includes are three parameters and their original cases, the other test has fixed parameters.

coleholcomb:
I refactored the nystrom attention test into two separate tests. The "close_to_sdp" test includes are three parameters and their original cases, the other test has fixed parameters.

I also returned the dropout = 0.0 cases, which has slightly increased the test suite runtime by about a minute in the buck test. Since some of these tests have many parameters (originally 2^9 test cases), it may be worthwhile to test the dropout code paths independently with new tests in the future.

Also want to note that there is another torch.allclose() with no assert statement on line 105 of test_compositional_attention.py. However, when I added the assert, the test failed even before I had changed the test parameters. May be worth examining.

tests/test_nystrom_attention.py
coleholcomb: By the way, is it intentional that the logic is repeated within the test_close_to_sdp() function?

blefaudeux:
> I refactored the nystrom attention test into two separate tests. The "close_to_sdp" test includes are three parameters and their original cases, the other test has fixed parameters.
> 
> I also returned the dropout = 0.0 cases, which has slightly increased the test suite runtime by about a minute in the buck test. Since some of these tests have many parameters (originally 2^9 test cases), it may be worthwhile to test the dropout code paths independently with new tests in the future.
> 
> Also want to note that there is another torch.allclose() with no assert statement on line 105 of test_compositional_attention.py. However, when I added the assert, the test failed even before I had changed the test parameters. May be worth examining.

thanks, I agree on the dropout case, maybe that it could be out of the combinatorics but tested on the side, that's also fine by me ! You can comment out the compositional_test then, it's on me, I'll file an issue right now

blefaudeux (approved):
LGTM, thanks again @coleholcomb !

PR #242, opened 2022-03-21
[minor] Speeding up comp attention test on CPU 

blefaudeux:
## What does this PR do?
required for FB infra, quick follow up from https://github.com/facebookresearch/xformers/pull/241, speed up the tests on cpu and in general 

## Before submitting

- [🙃] Did you have fun?
- [o] Did you make sure to update the docs?
- [o] Did you update the [changelog](https://github.com/facebookresearch/xformers/blob/master/CHANGELOG.md)? (if needed)

## PR review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -240,7 +240,7 @@ jobs:
       # Cache the venv directory that contains dependencies
       - restore_cache:
           keys:
-            - cache-key-cpu-py38-386-{{ checksum "requirements-test.txt" }}-{{ checksum ".circleci/config.yml" }}
+            - cache-key-cpu-py38-{{ checksum "requirements-test.txt" }}-{{ checksum ".circleci/config.yml" }}
 
       - <<: *install_dep
 
@@ -251,7 +251,7 @@ jobs:
       - save_cache:
           paths:
             - ~/venv
-          key: cache-key-cpu-py38-386-{{ checksum "requirements-test.txt" }}-{{ checksum ".circleci/config.yml" }}
+          key: cache-key-cpu-py38-{{ checksum "requirements-test.txt" }}-{{ checksum ".circleci/config.yml" }}
 
       - <<: *install_repo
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -240,7 +240,7 @@ jobs:
       # Cache the venv directory that contains dependencies
       - restore_cache:
           keys:
-            - cache-key-cpu-py38-386-{{ checksum "requirements-test.txt" }}-{{ checksum ".circleci/config.yml" }}
+            - cache-key-cpu-py38-{{ checksum "requirements-test.txt" }}-{{ checksum ".circleci/config.yml" }}
```
blefaudeux, line 243: I saw that on another branch, looks like the cached version is "poisoned" (installing torch nightly failed) and we keep pulling it. Ideally we should catch that, but this unlocks CI (generate a new cache, and -386 was an old reference anyway)

tests/test_compositional_attention.py
```diff
@@ -16,19 +16,18 @@
...
 
 
+@pytest.mark.parametrize("heads", [1, 4])
 @pytest.mark.parametrize("attn_dropout", [0.0, 0.3])
 @pytest.mark.parametrize("causal", [True, False])
-@pytest.mark.parametrize("heads", [1, 4])
 @pytest.mark.parametrize("rules", [1, 4])
 @pytest.mark.parametrize("q_compose", [False, True])
 @pytest.mark.parametrize("dim_selection", [MODEL // 2, None])
-@pytest.mark.parametrize("bias", [True, False])
```
blefaudeux, line 31: the bias part is not well tested per say, and it's just something we pass to the constructor (so it's tested / config typing already). Removing this halves the number of tests of course..

blefaudeux:
for this usecase the parameterize part is an overkill I think, in that most options don't interact with each other, so we don't really *need* to test all the combinations. The thing is that it's hard to be sure beforehand of what will never interact, so not completely trivial to prune the tree while staying on the safe side

PR #247, opened 2022-03-23
Add embedding projector for encoder and decoder

### LINKED ISSUE #244: Config does not support model_dim that does not match attention model_dim
# 🐛 Bug

Adding a `dim_model` within a `position_encoding_config` that is not equal to the `dim_model` of the rest of the model results in an unusable model, as the result of the embedding layer is no longer compatible with the feedforward  layers within the attention blocks.

This looks like a bug to me, since it doesn't ever allow you to specify a `dim_model` in the vocabulary that is not equal to the `dim_model` of the global model. In my use-case, in which I would like to tie the input of embeddings of two different models with different hidden sizes, it would be nice to have this work by inserting a linear layer in between the embedding layer and the first attention layer. 

## Command

Using the following config results in a crash:

```python
config = [{'reversible': False,
  'block_type': 'encoder',
  'num_layers': 4,
  'dim_model': 64,
  'layer_norm_style': 'pre',
  'position_encoding_config': {'name': 'vocab',
   'seq_len': 512,
   'vocab_size': 30522,
   'dim_model': 256},
  'multi_head_config': {'num_heads': 1,
   'residual_dropout': 0,
   'use_rotary_embeddings': True,
   'attention': {'name': 'scaled_dot_product',
    'dropout': 0.1,
    'causal': False,
    'seq_len': 512,
    'num_rules': 4}},
  'feedforward_config': {'name': 'MLP',
   'dropout': 0.1,
   'activation': 'gelu',
   'hidden_layer_multiplier': 4}}]

import torch
from xformers.factory.model_factory import xFormer, xFormerConfig

cfg = xFormerConfig(config)
model = xFormer.from_config(cfg)

X = torch.randint(0, 1000, (10, 256))

model(X)
```

```
RuntimeError: Given normalized_shape=[64], expected input with shape [*, 64], but got input of size[10, 256, 25
6]
```

## To Reproduce

See above

## Expected behavior

I'd expect the model builder to complain that adding a `dim_model` within a vocab block does not make sense, or I expect it to work.

stephantul:
## What does this PR do?
Fixes #244 .

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [x] N/A
  - [x] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [0.0.x] - TBD
+### Fixed
+- Fix some torchscriptability [#246]
 
 ## [0.0.10] - 2022-03-14
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Expose bias flag for feedforwards, same default as Timm [#220]
 - Update eps value for layernormm, same default as torch [#221]
 - PreNorm bugfix, only one input was normalized [#233]
+- Fix bug where embedding dimensions that did not match model dim would lead to a crash [#244]
 
 ### Added
 - Add DeepNet (DeepNorm) residual path and init [#227]
--- a/xformers/factory/block_factory.py
+++ b/xformers/factory/block_factory.py
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 
+import logging
 from dataclasses import asdict, dataclass
 from enum import Enum
 from typing import Any, Dict, Optional, Tuple, Union
@@ -39,7 +40,7 @@ class LayerPositionBitmask(int, Enum):
 
 
 class LayerPosition:
-    """ Bitmask to mark this layer as first, last, nothing or both"""
+    """Bitmask to mark this layer as first, last, nothing or both"""
 
     def __init__(self):
         self.bitmask = LayerPositionBitmask.Default
@@ -154,7 +155,8 @@ class xFormerBlockConfig:
 
         # Fill in possible gaps in the config for subparts of the block
         self.feedforward_config = generate_matching_config(
-            feedforward_config, FEEDFORWARD_REGISTRY[feedforward_config["name"]].config,
+            feedforward_config,
+            FEEDFORWARD_REGISTRY[feedforward_config["name"]].config,
         )
 
         self.position_encoding_config = (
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Expose bias flag for feedforwards, same default as Timm [#220]
 - Update eps value for layernormm, same default as torch [#221]
 - PreNorm bugfix, only one input was normalized [#233]
+- Fix bug where embedding dimensions that did not match model dim would lead to a crash [#244]
```
blefaudeux, line 14: this would need to be moved up, it would be part of the next release ? I'm kind of on the fence as of whether this is a "bug" (there are plenty of settings which you can pass which will "crash", in that they will tell you down the line that the dimensions mismatch), but as you see fit

xformers/factory/block_factory.py
```diff
@@ -303,6 +302,13 @@ def __init__(self, config: xFormerEncoderConfig, **kwargs):
             else None
         )
 
+        if self.pose_encoding is not None:
+            pos_encoding_dim = config.position_encoding_config.dim_model
+            mha_dim = config.multi_head_config["dim_model"]
+
+            if pos_encoding_dim != mha_dim:
+                self.embedding_projector = nn.Linear(pos_encoding_dim, mha_dim)
```
blefaudeux, line 310: could you add a warning here ? (logging.warning) It will be only visible at construction time, so no spamming, but could make it easier to debug if this is not on purpose

blefaudeux:
looks good to me except for minor remarks, thanks a lot @stephantul ! I can fix this and the linting if you'd like, else feel free to update the PR anytime

CHANGELOG.md
stephantul: Ah yeah, not a bug, sorry! Feel free to rename it as you see fit.

stephantul:
I'm not sure why black is throwing an error here, but not locally. I checked before pushing, but black is still saying everything is fine. So if you could fix that, that would be best.

blefaudeux:
@stephantul see https://github.com/stephantul/xformers/pull/1 for a proposed fix

stephantul:
Thanks for the fix! I was offline during the weekend, and was planning to fix it today, but I'm glad you took care of it 😄

blefaudeux:
LGTM @stephantul, thank you ! One last thing would be to unit test this case, it can be just a matter of adding another case [here](https://github.com/facebookresearch/xformers/blob/main/tests/test_embedding.py#L23). I can do that later on if you prefer, I understand that there are a lot of steps to contributing and wouldn't want to wear you down :)

blefaudeux (approved):
LGTM, as you prefer unit test wise, I can follow up on that ! Thanks a bunch @stephantul

blefaudeux:
landing that, adding more unit tests in a seperate PR, easy follow up

PR #266, opened 2022-04-09
[feat][minor] 2/3 Make it explicit whether an attention mechanism supports a mask

blefaudeux:
## What does this PR do?
Preamble to the Triton2 PR (since triton 2 will change the blocksparse attention and not support attention masks anymore)

The unit test which now fails on CI should be unrelated to this PR, and is fixed by the next PR in line, triton2

## Before submitting
- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/tests/test_model_factory.py
+++ b/tests/test_model_factory.py
@@ -39,7 +39,7 @@ encoder_configs = {
         "num_heads": 4,
         "residual_dropout": 0,
         "attention": {
-            "name": "linformer",
+            "name": "scaled_dot_product",
             "dropout": 0,
             "causal": True,
             "seq_len": SEQ,
@@ -73,7 +73,7 @@ decoder_configs = {
         "residual_dropout": 0,
         "dim_model": EMB,
         "attention": {
-            "name": "linformer",
+            "name": "scaled_dot_product",
             "dropout": 0,
             "causal": True,
             "seq_len": SEQ,
@@ -84,7 +84,7 @@ decoder_configs = {
         "residual_dropout": 0,
         "dim_model": EMB,
         "attention": {
-            "name": "linformer",
+            "name": "scaled_dot_product",
             "dropout": 0,
             "causal": True,
             "seq_len": SEQ,
--- a/xformers/components/attention/base.py
+++ b/xformers/components/attention/base.py
@@ -53,6 +53,10 @@ class Attention(nn.Module, metaclass=ABCMeta):
         # so that the MHA wrapper should skip it
         self.requires_skip_multi_head = False
 
+        # Whether this attention mechanism supports attention masks
+        self.supports_attention_mask = True
+        self.supports_key_padding_mask = False
+
     @classmethod
     def from_config(cls: Type[Self], config: AttentionConfig) -> Self:
         # Generate the class inputs from the config
--- a/tests/test_block_factory.py
+++ b/tests/test_block_factory.py
@@ -50,6 +50,7 @@ def test_xformer_encoder_block(
     device: torch.device,
     reversible: bool,
 ):
+
     block_size = 16
 
     attention_config = {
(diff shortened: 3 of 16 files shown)


.circleci/config.yml
```diff
@@ -150,7 +150,7 @@ run_unittests: &run_unittests
   - run:
       name: Run Unit Tests
       command: |
-        pytest --junitxml=test-results/junit.xml --verbose --timeout 600 tests
+        CUDA_LAUNCH_BLOCKING=1 pytest --junitxml=test-results/junit.xml --verbose --timeout 600 tests
```
blefaudeux, line 153: @fmassa CI crash happens with or without this, I added that to make sure that it was not another test before which was silently failing. The crash is in a sputnik kernel, but it makes little sense to me given the contents of this PR

tests/test_model_factory.py
```diff
@@ -39,7 +39,7 @@
         "num_heads": 4,
         "residual_dropout": 0,
         "attention": {
-            "name": "linformer",
+            "name": "scaled_dot_product",
```
blefaudeux, line 42: we were actually passing an attention mask in this test, which makes little sense with linformer. Now that we assert in this case, it had to be fixed :)

blefaudeux:
@dianaml0 @finassa this goes with the triton2 PR again

xformers/components/attention/base.py
```diff
@@ -53,6 +53,10 @@ def __init__(self, dropout: Optional[float] = None, *args, **kwargs):
         # so that the MHA wrapper should skip it
         self.requires_skip_multi_head = False
 
+        # Whether this attention mechanism supports attention masks
+        self.supports_attention_mask = True
+        self.supports_key_padding_mask = False
```
fmassa, line 58: Ideally, I think I would prefer if we only support `attention_mask`, and then support the "optimized" case of `key_padding_mask` internally by checking the strides of the `attention_mask` Tensor.
The `key_padding_mask` case is actually just assuming that the stride of dimension -2 of the `attention_mask` is 0. So the converting `key_padding_mask` to a `attention_mask` can be done efficiently via
```python
attn_mask = key_padding_mask[:, None, None, :].expand(batch, heads, query_len, key_len)
```

Indeed, are we sure that all our implementations support both `attention_mask` *and* `key_padding_mask` at the same time?

fmassa (approved):
LGTM, thanks!

I have only one comment (which is sort of unrelated to this PR), but which I'd love to get your opinion on.

xformers/components/attention/base.py
blefaudeux: Agree on the simpler interface, it's a bit confusing right now. One trick is that for some attentions (Nystrom for instance) there's no attention mask but there's a possible key padding mask (which can then generate a mask of sorts indeed, but in that case the dimensions will differ). I think we could do a second pass on that indeed, thoughts @dianaml0  ?

dianaml0: I agree ideally we could have this simpler approach but I ended up adding a key padding mask argument separately for Nystrom since Nystrom can't apply attention masks. I couldn't come up with a way around that but maybe there's a better way

dianaml0 (approved):

xformers/components/attention/base.py
blefaudeux: hmm so ideally we would just support key padding for nystrom and not make it visible everywhere ? right now there's a global flag because this happens in the MHA, but maybe that we could more this one level down and not expose this for other mechanisms ?

blefaudeux: else it feels like an abstraction leak on our end, not perfect.. could be worth some more thinking

PR #271, opened 2022-04-13
[RFC] 1/3 Moving the CI to conda, picking a more modern cuda + pytorch combo

blefaudeux:
## What does this PR do?
Not super pretty, suggestions welcome. The crux of the matter was that circleCI supports images with cuda 11.1 (soon to be deprecated) and 11.4, while pytorch nightlies are built for everything but cuda 11.4..

*a* (note the *a*, unicity is not proven) solution is to rely on conda instead, which handles both pytorch and the matching cuda in a pinch. The handling of shells in circleci is a semi-mistery to me, but in short most of the conda mechanics do not work (the shells have no .bashrc for instance), and I changed the scripts to point to the conda installed python instead (after many more elegant tests which did not work)

A Triton unit test segfaults now, but this is with an old version of triton, I think that spending time on that is ill-advised if we can agree on the next-next-PR (triton2) and land all of them instead

## Before submitting

- [x] Did you have fun?

## PR review


blefaudeux:
@fmassa @dianaml this goes with the triton2 PR, the plan was to land all three in one go

fmassa (approved):
LGTM, thanks a lot!

Another approach we could follow in the future would be to use the docker images from PyTorch, which comes with a number of [things already packed in](https://github.com/pytorch/vision/blob/3122ea16057ebcc748d4743055e84c60a1805131/.circleci/config.yml.in#L786-L791)

That being said, the PyTorch team will at some point in the future provide a set of helper functions so that dealing with all of this would be simpler, so we will probably just use that when they get available.

blefaudeux:
> LGTM, thanks a lot!
> 
> 
> 
> Another approach we could follow in the future would be to use the docker images from PyTorch, which comes with a number of [things already packed in](https://github.com/pytorch/vision/blob/3122ea16057ebcc748d4743055e84c60a1805131/.circleci/config.yml.in#L786-L791)
> 
> 
> 
> That being said, the PyTorch team will at some point in the future provide a set of helper functions so that dealing with all of this would be simpler, so we will probably just use that when they get available.

Oh that would be great for the docker image ! I had a quick look and it seemed that we had to setup the actual image hosting on top, so I skipped that for now, but if there's an existing image hosted by pytorch that would be perfect 😍

.circleci/config.yml
```diff
@@ -15,12 +15,12 @@ cpu_py38: &cpu_py38
     - image: cimg/python:3.8
   resource_class: large
 
-gpu_cu111: &gpu_cu111
+gpu_cu114: &gpu_cu114
   environment:
-    CUDA_VERSION: "11.1"
-    CUDA_HOME: /usr/local/cuda-11.1
+    CUDA_VERSION: "11.4"
```
dianaml0, line 20: How come 11.4 and not a later one?

dianaml0 (approved):
This is great, thanks! :)

PR #263, opened 2022-04-07
Added SmeLU

kashif:
## What does this PR do?
Fixes #262 .

## Before submitting

- [x] Did you have fun?
  - Make sure you had fun coding 🙃
- [x] Did you read the [contributor guideline](https://github.com/facebookresearch/xformers/blob/master/CONTRIBUTING.md)?
- [ ] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
  - [x] N/A
- [ ] Did you make sure to update the docs?
  - [ ] N/A
- [ ] Did you write any new necessary tests?
  - [ ] N/A
- [ ] Did you update the [changelog](https://github.com/facebookresearch/xformers/blob/master/CHANGELOG.md)? (if needed)
  - [ ] N/A

## PR review
Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

PR #272, opened 2022-04-13
[backend] 3/3 Triton 2 update

blefaudeux:
## What does this PR do?
Push things a little forward with triton 2. My thinking was to try to land all 3 PRs in one go when this last one is green
Happy to update this PR on that front (and others), this is something quickly wrapped together to try to unlock some other PRs (like #263)

TODOs:
- [x] Upgrade the CI
- [x] Fix the kernel syntax to triton 2
- [x] Fix the timm benchmark / points to layernorm being broken
- [x] Sparse softmax / bmm unit test failure -> triton2 changed some of the internal computation formats (and pytorch did the same, with a default switch to tf32 for some matmuls I believe). @fmassa the sparse_tensor unit tests were not passing even if almost nothing in the codebase had changed  (except for newer CUDA + Triton2), so this PR relaxed some of the parity constraints. Please have a look, we can change things
- [x] Some speed improvements given triton2 changes

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review


blefaudeux:
to peek into the remaining issues: https://app.circleci.com/pipelines/github/facebookresearch/xformers/1395/workflows/b90368b4-eda1-4037-8cd3-a4d138b6e320/jobs/3075 getting there

blefaudeux:
@dianaml0 just FYI, investigating the layernorm crash, this works fine with cuda 11.6 / Ampere actually, but I can repro with 11.4

blefaudeux:
> @dianaml0 just FYI, investigating the layernorm crash, this works fine with cuda 11.6 / Ampere actually, but I can repro with 11.4

some writes were not masked, could be why. I just pushed a small update, works on my machine (tm)

blefaudeux:
> @dianaml0 just FYI, investigating the layernorm crash, this works fine with cuda 11.6 / Ampere actually, but I can repro with 11.4

layernorm fixed, as far as I can see, should be a bit faster for very big sizes also

blefaudeux:
@fmassa if you have some cycles at some point, would you mind having a look at the test_sparse_softmax test which does not pass with this branch ? You have a little more context around there (and you could check the changes from this PR, probably a bit rough in this area)

blefaudeux:
> @fmassa if you have some cycles at some point, would you mind having a look at the test_sparse_softmax test which does not pass with this branch ? You have a little more context around there (and you could check the changes from this PR, probably a bit rough in this area)

I just improved a bit on the changes in /sparse_tensor, should be ok by now except r the fact that the new blocksparse does not acccept a per pixel mask anymore, so this probably breaks some of this abstraction. Something else which is a little strange is that with the updated cuda *both* sputnik sparse and blocksparse fail the test_sparse_tensors unit test, despite almost no changes there from this PR. Blocksparse makes sense to me with the mask change I just mentioned, sparse is a
sparse
 little perplexing. No obvious big pattern in the error map 

 blocksparse (makes sense, probably because of the missing mask)
blocksparse

(per pixel sparse, not sure why this would break)
sparse

tests/test_sparse_tensors.py
```diff
@@ -56,6 +56,29 @@ def _create_tensor(tensor_type, device, dtype, shape, sparsity):
...
+    if tensor_type == BlockSparseTensor and "cuda" in device:
+        # Upstream GPU blocksparse (Triton op) uses TF32 by default for all internal computations
+        # TF32 has the precision of fp16 but the range of fp32
+        # See https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format/
+        torch.backends.cuda.matmul.allow_tf32 = True
+        torch.backends.cudnn.allow_tf32 = True
+        return torch.float32, 1e-1
+
+    # Force pytorch to keep its computations as float32 (will default to tf32 with recent cuda and ampere+ GPU)
+    torch.backends.cuda.matmul.allow_tf32 = False
```
blefaudeux, line 76: @fmassa this fixed issues that I was seeing with these unit tests on an ampere GPU, which I presume stemmed from the fact that the sparse kernels were fp32 while pytorch defaulted to tf32

tests/test_sparse_tensors.py
```diff
@@ -56,6 +56,29 @@ def _create_tensor(tensor_type, device, dtype, shape, sparsity):
...
+
+
+def _get_dtype_atol(tensor_type, device: str):
+    _seed()
+
+    if tensor_type == BlockSparseTensor and "cuda" in device:
+        # Upstream GPU blocksparse (Triton op) uses TF32 by default for all internal computations
+        # TF32 has the precision of fp16 but the range of fp32
+        # See https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format/
+        torch.backends.cuda.matmul.allow_tf32 = True
```
blefaudeux, line 71: @fmassa this seems to be required following the switch to triton2, which internally moved all tl.dot() operations to tf32

tests/test_sparse_tensors.py
```diff
@@ -56,6 +56,29 @@ def _create_tensor(tensor_type, device, dtype, shape, sparsity):
         )
 
 
+def _seed():
+    torch.random.manual_seed(42)
+    torch.cuda.manual_seed_all(42)
+
+
+def _get_dtype_atol(tensor_type, device: str):
+    _seed()
```
blefaudeux, line 65: this was to remove some reproducibility issues in between circleci and my machine..

tests/test_triton_blocksparse.py
```diff
@@ -72,7 +72,14 @@ def test_matmul(MODE, TRANS_A, TRANS_B, BLOCK, DTYPE, Z=32, H=2, M=512, N=384, K
...
 
     # triton result
-    op = blocksparse_matmul(layout, BLOCK, MODE, trans_a=TRANS_A, trans_b=TRANS_B)
+    op = blocksparse_matmul(
+        layout,
+        BLOCK,
+        MODE,
+        trans_a=TRANS_A,
+        trans_b=TRANS_B,
+        device=torch.device("cuda"),
```
blefaudeux, line 81: triton blocksparse op now requires the device to be passed in

tests/test_triton_blocksparse.py
```diff
@@ -103,34 +110,15 @@ def test_softmax(BLOCK, WIDTH, DTYPE):
...
-        low=0, high=2, size=(N, N), dtype=torch.bool, requires_grad=False, device="cuda"
-    )
-    kp_mask = torch.randint(
-        low=0, high=2, size=(Z, N), dtype=DTYPE, requires_grad=False, device="cuda"
-    )
-    kp_mask[kp_mask == 1.0] = float("-inf")
 
     # triton result
-    op = blocksparse_softmax(layout, BLOCK)
+    op = blocksparse_softmax(layout, BLOCK, device=torch.device("cuda"))
```
blefaudeux, line 115: triton blocksparse softmax now requires the device to be passed in

tests/test_triton_blocksparse.py
```diff
@@ -103,34 +110,15 @@ def test_softmax(BLOCK, WIDTH, DTYPE):
...
     tx = block_sparsify_tensor(x, layout, BLOCK)
-    ty = op(
-        tx,
-        scale=scale,
-        key_padding_mask=kp_mask,
```
blefaudeux, line 120: triton blocksparse now does not support attn mask or key_padding mask. We can pass a "causal" flag though

tests/test_triton_dropout.py
```diff
@@ -88,14 +88,16 @@ def test_dropout(shape, amp, bias, p):
         x_ref = (x + b if bias else x).to(y.dtype)
         assert torch.allclose(x_ref, y, rtol=tol), f"{x[x>y]}"
 
-        # Check that 1 means dropout for sure
-        y = triton_dropout(x, p=1, bias=b)
```
blefaudeux, line 92: minor change, p=1 means nothing out, which does not make a lot of sense (technically we should skip the kernel and return torch.zeros_like()..)

tests/test_triton_layernorm.py
```diff
@@ -50,7 +50,7 @@ def test_layernorm_parity(shape, amp):
     torch.random.manual_seed(0)
     X_ = torch.normal(0, 1, size=shape, device="cuda", requires_grad=True)
 
-    eps = 1e-5
+    eps = 1e-4
```
blefaudeux, line 53: 1/1e-5 overflows in fp16..

xformers/__init__.py
```diff
@@ -8,7 +8,7 @@
 import torch
 
 # Please update the doc version in docs/source/conf.py as well.
-__version__ = "0.0.10"
+__version__ = "0.0.11.dev"
```
blefaudeux, line 11: nit, but I think that it's nice if our current main is not numbered like the pip package, so that we can dissociate in between the two

xformers/components/attention/blocksparse.py
```diff
@@ -86,39 +85,24 @@ def __init__(
             ), "Only block sizes in [16, 32, 64] are supported"
 
             super().__init__()
+
+            self.causal = causal
```
blefaudeux, line 89: triton blocksparse can use this

xformers/components/attention/blocksparse.py
```diff
@@ -86,39 +85,24 @@ def __init__(
...
             self.requires_head_dimension = True
 
             # key padding mask and attention mask must be passed in separately
-            self.requires_separate_masks = True
             self.requires_same_k_q_dimensions = True
 
-            # Properties specific to this attention mechanism
-            self.supports_attention_mask = True
-            self.supports_key_padding_mask = True
+            # The underlying triton op does not support per element attention mask
```
blefaudeux, line 103: see previous PR which introduced these flags, we can now flip them in this case

xformers/components/attention/blocksparse.py
```diff
@@ -133,33 +117,46 @@ def forward(
...
+            .. note: Per element attention mask is not supported, but you can specify causality
             """
 
-            # NOTE:
-            # The attention mask will be taken into account when computing the softmax
-            # meaning that non-masked values which are present in the initial blocksparse layout will be computed.
-            # If blocks are to be constantly masked, better perf would thus be reached by signalling them out in the
-            # initial attention setup
+            # Delayed triton init, to make sure that we get the right device
+            if not hasattr(self, "sparse_dot_sdd"):
```
blefaudeux, line 135: triton blocksparse need the correct device to be passed, but it could be that it's not the case at construction time (if constructed on CPU then moved), and it's not possible to just default to cuda:0 (would break many multi-gpu cases). So we defer the construction until after the first input tensor comes in

blefaudeux:
I think that there's some margin in terms of speed around the fused linear layers, the matmul triton op improved a lot in the last few months and fused linear could probably benefit from some of it. Better keeping it for another PR though

colehawkins:
@blefaudeux larger block sizes (at least up to 128) are supported by triton v2. 

Also, is any type of recompile option possible for the blocksparse attention? I ran into this issue using tritonv2 blocksparse attention in a distributed environment (specifically huggingface trainer). Since the ops are device-specific, they need to be recreated or there is a device mismatch. It's not too hard to work around with a delayed device-specific model initialization, but I think there are potentially smoother work arounds by inheriting the `to` function or attempting re-creation in the event of a device mismatch. It's also possible that I'm missing a trivial workaround.

Happy to submit both in a PR to either branch, much more timely than last time (1-2 days).

blefaudeux:
> @blefaudeux larger block sizes (at least up to 128) are supported by triton v2. 
> 
> 
> 
> Also, is any type of recompile option possible for the blocksparse attention? I ran into this issue using tritonv2 blocksparse attention in a distributed environment (specifically huggingface trainer). Since the ops are device-specific, they need to be recreated or there is a device mismatch. It's not too hard to work around with a delayed device-specific model initialization, but I think there are potentially smoother work arounds by inheriting the `to` function or attempting re-creation in the event of a device mismatch. It's also possible that I'm missing a trivial workaround.
> 
> 
> 
> Happy to submit both in a PR to either branch, much more timely than last time (1-2 days).

Oh sure for the block size, I didn't know, and there may be a type check or cast which should be removed also (top of head we used to force cast to fp16).

For the device, overloading .to() is possible but there can be a lot of cases to handle (for instance if blocksparse is part of a wrap which intercepts this call), I'm not sure that it will be much cleaner ? The first FW is much slower with Triton anyway due to the JIT, so I think that this has no measurable perf impact. If .to() can read cleaner then sure, and PR welcome to this branch if you want ? 

No problem for the timing, not very timely myself on the topic :)

xformers/components/attention/blocksparse.py
```diff
@@ -86,39 +85,24 @@ def __init__(
...
             self.requires_same_k_q_dimensions = True
 
-            # Properties specific to this attention mechanism
-            self.supports_attention_mask = True
-            self.supports_key_padding_mask = True
+            # The underlying triton op does not support per element attention mask
+            self.supports_attention_mask = False
+            self.supports_key_padding_mask = False
 
         def update_mask_type(self, mask: torch.Tensor):
```
SeanNaren, line 107: Was just reading the code and realised this is not used anymore, safe to delete?

blefaudeux: Nice catch ! In general I need to give it a second look and clean things up, I was waiting for @colehawkins so that there's no conflict but will do

colehawkins: @blefaudeux  Posted in #277. Pending CI, but tests passed locally so I have high hopes.

tests/test_sparse_tensors.py
fmassa: oh wow, thanks for spotting this!

One more instance where tf32 is being somewhat harmful. Maybe worth commenting on https://github.com/pytorch/pytorch/issues/67384  ?

tests/test_sparse_tensors.py
fmassa: SGTM wrt the tests!

tests/test_sparse_tensors.py
```diff
@@ -56,6 +56,29 @@ def _create_tensor(tensor_type, device, dtype, shape, sparsity):
...
+
+    if tensor_type == BlockSparseTensor and "cuda" in device:
+        # Upstream GPU blocksparse (Triton op) uses TF32 by default for all internal computations
+        # TF32 has the precision of fp16 but the range of fp32
+        # See https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format/
+        torch.backends.cuda.matmul.allow_tf32 = True
+        torch.backends.cudnn.allow_tf32 = True
+        return torch.float32, 1e-1
```
fmassa, line 73: wow, that is quite some low precision...

xformers/components/attention/blocksparse.py
fmassa: sounds good to me.

One other option would be to inherit the `.cuda()` / `.to()` methods so that they re-create those objects.

The current approach is fine as is because the those objects don't contain learnable parameters, but if that was the case it would mess up with optimizers / distributed.

xformers/components/attention/blocksparse.py
fmassa: This is in principle a BC-breaking change. Should we bother, or should we just follow what Triton does?

xformers/sparse/blocksparse_tensor.py
```diff
@@ -195,9 +201,7 @@ def _softmax(cls, arg0, dim):
         if not (dim == -1 or dim == 2):
             return NotImplemented
         if _can_use_triton(arg0):
-            # TODO triton softmax performs an in-place operation
-            # res = arg0.__sparse_softmax(arg0.__values)
-            res = arg0.__sparse_softmax(arg0.__values.clone())
+            res = arg0.__sparse_softmax(arg0.__values)
```
fmassa, line 204: nice!

xformers/triton/dropout.py
```diff
@@ -166,15 +169,22 @@ def dropout(
     Optionally add a bias, the computation will be fused.
     """
 
+    assert p < 1.0, f"We don't want to drop all the values, most probably {p}"
```
fmassa, line 172: PyTorch supports this case, so if we want our dropout to be a drop-in replacement to PyTorch's implementation it would be good to support this as well.

fmassa (approved):
Thanks for the PR!

I've left a few comments, but given the size of this PR already, I'd propose that we address them in a separate PR after this one is merged.

tests/test_sparse_tensors.py
blefaudeux: It's a strange format, range of fp32 but precision of fp16, it's also kind of peculiar that it's really 18bits but named tf32..

blefaudeux:
checking right now with a small ViT/cifar training, then landing the whole stack asap

xformers/components/attention/blocksparse.py
blefaudeux: basically we don't have much choice, short of implementing blocksparse ourselves ? Phil arguments were that it was not typically used (short of causal, which is supported), and that the attention mask (additive -> floats) ended up taking a significant amount of space in memory. My take would be that we have a in-house fallback since people can use sparse attention, and else they can stick to the current pip release for some time, so not something that I would do every day (breaking BC), but in that case that was ok ?

xformers/components/attention/blocksparse.py
blefaudeux: yep, that's something we discussed with @colehawkins on another PR, one possible issue for me is that some sharded trainers intercept the .to() calls, so this would silently fail in that case. I think that both takes have issues (the delayed init and the .to() overload), the only clean way out that I can think of is to make this attention take a "device" as a construction argument, and put it to the right place from the beginning

xformers/triton/dropout.py
blefaudeux: fixed, incoming update, thanks for the catch !

tests/test_triton_dropout.py
blefaudeux: fixed, I brought both the test and implementation back

blefaudeux:
I tried to address most comments @fmassa, we can follow up on the version numbering and delayed init in another PR. I just checked with my "classical" ViT/Cifar test, just in case, same accuracy as before when pulling in all the triton layers

xformers/components/attention/blocksparse.py
colehawkins: @blefaudeux @fmassa  I've been using this with both pytorch lightning and the huggingface trainer, and this method (initialization at first forward) is the cleanest way I found that doesn't break any standard workflows or require model initialization workarounds. If we just take the device at construction this breaks the huggingface trainer "natural approach" for single-node, multi-gpu which (1) create the model, then (2) call `.to()`.

One alternative is to initialize with a device at construction, keep that as `self.device`, and then check that against `query.device` and possibly re-initialize at the forward pass.

PR #278, opened 2022-04-21
Remove keyword arguments to attention call, pass as positional

SeanNaren:
## What does this PR do?

As discussed with Ben offline, when using forward module hooks on the attention block arguments need to be passed in as positional args to access them within the hook (can be seen in the [docs](https://pytorch.org/docs/stable/generated/torch.nn.modules.module.register_module_forward_hook.html#torch.nn.modules.module.register_module_forward_hook)).

My use case for this is profiling using the [DeepSpeedProfiler](https://www.deepspeed.ai/tutorials/flops-profiler/). The profiler attaches hooks to modules to calculate how many FLOPs are done. When using xformers we need to declare a few more hooks in order to calculate them correctly, (as seen [here](https://github.com/SeanNaren/SmallScience/blob/feat/profiling/profiler.py#L26)).

An example!
```python
import torch

from xformers.components.attention import ScaledDotProduct

attention = ScaledDotProduct()


def hook(module, input, output):
    assert input


attention.register_forward_hook(hook)

inputs = torch.rand((8, 128, 128))

# this would fail the above assertion
# attention(q=inputs, k=inputs, v=inputs)

# this works!
attention(inputs, inputs, inputs)
```

@blefaudeux mentioned how it was mainly a case of being explicit (as it could be very wrong if args were swapped, the attention mechanism under the hood expected a different order of positional arguments). I'm up for alternatives, please let me know if there are any!

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/xformers/components/multi_head_dispatch.py
+++ b/xformers/components/multi_head_dispatch.py
@@ -210,7 +210,7 @@ class MultiHeadDispatch(nn.Module):
             v = reshape_fn(v, B, S_K, self.num_heads, self.dim_k)
 
         # Self-attend
-        y = self.attention(q=q, k=k, v=v, **kw_mask_args)
+        y = self.attention(q, k, v, **kw_mask_args)
 
         # Re-assemble all head outputs side by side
         y = (


xformers/components/multi_head_dispatch.py
```diff
@@ -210,7 +210,7 @@ def check(t, name):
             v = reshape_fn(v, B, S_K, self.num_heads, self.dim_k)
 
         # Self-attend
-        y = self.attention(q=q, k=k, v=v, **kw_mask_args)
+        y = self.attention(q, k, v, **kw_mask_args)
```
blefaudeux, line 213: I think that works, since all the mechanisms need to inherit from the base class, and the base class uses qkv ([see](https://github.com/facebookresearch/xformers/blob/main/xformers/components/attention/base.py#L71)). If it was not for the downstream issue for you I would have kept the keywords, but I think that it's a good enough reason, xformers does not exist in a vacuum and being interoperable with the AI ecosystem is worth something..

blefaudeux (approved):
works for me, the base class enforces this ordering so it should catch obvious mistakes. @fmassa @dianaml0 feel free to jump in and disagree :)

blefaudeux:
(Checked by Diana, as far as I understood :))

SeanNaren:
Thanks @blefaudeux and co :)

Issue #161, opened 2021-12-21
[feat] Add a fast implementation of Rabe and Staats algorigthm (mem efficient attention) on GPU


blefaudeux:
# 🚀 Feature
Implement https://arxiv.org/pdf/2112.05682v2.pdf using Triton

## Motivation
There are [existing implementations](https://github.com/aqlaboratory/openfold/commit/96709588dd6759a02a2eca8b67a278bd3219a3f9) in Pytorch, but they re bound to be a little slow. It s actually not that much work to write that down in Triton, give it a shot. Given the FW speed (should be similar to normal attention, without the memory) and the expected BW speed (about 60% of the vanilla attention), feels like a compromise that many would use

## Pitch
The required kernel is actually not that far from some of the kernels that we already have, at least for the FW. The  chunk strategy proposed by the paper is actually fairly classic in that field, nothing out of the ordinary ([see for instance](https://triton-lang.org/getting-started/tutorials/03-matrix-multiplication.html#sphx-glr-getting-started-tutorials-03-matrix-multiplication-py)), so it's bound to be pretty fast if correctly implemented.

## Alternatives
At least support a pure pytorch variant in xformers ?

erip:
Another reference impl can be found [here](https://github.com/AminRezaei0x443/memory-efficient-attention) -- same caveats as outlined above.

blefaudeux:
I've started something, it feels like some of the logic would need to be changed a bit for that to make sense at a kernel level, at least for triton. In particular it's hard to sequence things outside of a kernel, and reproducing the same logic as the one from the paper would lead to big buffers (if the computation is tiled), which diminish the interest a lot. It feels like the best approach is with a kernel owning the whole line, and a couple of rows at a time to help with data fetch reuse

blefaudeux:
> Another reference impl can be found [here](https://github.com/AminRezaei0x443/memory-efficient-attention/blob/main/memory_efficient_attention/attention_torch.py) -- same caveats as outlined above.

thanks !

PR #281, opened 2022-04-22
Memory efficient attention - backward pass

fmassa:
## What does this PR do?

This PR implements the memory-efficient attention mechanism from https://arxiv.org/pdf/2112.05682v2.pdf, with both CPU and CUDA kernels, targetting the backward implementation. For now, only fp32 is supported.

The CPU implementation is naive and not meant to be fast, and is there only as a reference.

The CUDA implementation has competitive runtimes compared to a vanilla PyTorch implementation, while using 10x+ less memory. 

Contrary to the forward implementation, the backwards supports inputs of arbitrary number of embedding sizes. I'll probably update the forward implementation to use a similar approach in the future.

In order to keep the backwards somewhat efficient, we need to return the `logsumexp` during forward as well. For some reason, I needed to template it in CUDA otherwise I would face performance slowdowns.

In the same vein, in order to support arbitrary sequence lengths, I use a masking approach. But the masking (and in particular the `min(index, M)` operations) are very slow, so I templated this as well so that we don't need to run this if the sequence length is a multiple of 32.

<details>

<summary>
Full details for the benchmark I run can be found in here
</summary>

```
Backward
===== (1, 127, 16) =====
Optimized Memory used: 0.07861328125 MB
Vanilla Memory used: 0.310546875 MB
===== (1, 127, 32) =====
Optimized Memory used: 0.15673828125 MB
Vanilla Memory used: 0.373046875 MB
===== (1, 128, 16) =====
Optimized Memory used: 0.07861328125 MB
Vanilla Memory used: 0.3125 MB
===== (1, 128, 32) =====
Optimized Memory used: 0.15673828125 MB
Vanilla Memory used: 0.375 MB
===== (1, 512, 16) =====
Optimized Memory used: 0.314453125 MB
Vanilla Memory used: 4.25 MB
===== (1, 512, 32) =====
Optimized Memory used: 0.626953125 MB
Vanilla Memory used: 4.5 MB
===== (1, 513, 16) =====
Optimized Memory used: 0.31982421875 MB
Vanilla Memory used: 4.271484375 MB
===== (1, 513, 32) =====
Optimized Memory used: 0.63232421875 MB
Vanilla Memory used: 4.521484375 MB
===== (1, 1023, 16) =====
Optimized Memory used: 0.62890625 MB
Vanilla Memory used: 16.470703125 MB
===== (1, 1023, 32) =====
Optimized Memory used: 1.25390625 MB
Vanilla Memory used: 16.970703125 MB
===== (1, 1024, 16) =====
Optimized Memory used: 0.62890625 MB
Vanilla Memory used: 16.5 MB
===== (1, 1024, 32) =====
Optimized Memory used: 1.25390625 MB
Vanilla Memory used: 17.0 MB
===== (8, 127, 16) =====
Optimized Memory used: 0.6240234375 MB
Vanilla Memory used: 2.466796875 MB
===== (8, 127, 32) =====
Optimized Memory used: 1.244140625 MB
Vanilla Memory used: 2.962890625 MB
===== (8, 128, 16) =====
Optimized Memory used: 0.62890625 MB
Vanilla Memory used: 2.5 MB
===== (8, 128, 32) =====
Optimized Memory used: 1.25390625 MB
Vanilla Memory used: 3.0 MB
===== (8, 512, 16) =====
Optimized Memory used: 2.515625 MB
Vanilla Memory used: 34.0 MB
===== (8, 512, 32) =====
Optimized Memory used: 5.015625 MB
Vanilla Memory used: 36.0 MB
===== (8, 513, 16) =====
Optimized Memory used: 2.52099609375 MB
Vanilla Memory used: 34.130859375 MB
===== (8, 513, 32) =====
Optimized Memory used: 5.02587890625 MB
Vanilla Memory used: 36.134765625 MB
===== (8, 1023, 16) =====
Optimized Memory used: 5.0263671875 MB
Vanilla Memory used: 131.99609375 MB
===== (8, 1023, 32) =====
Optimized Memory used: 10.021484375 MB
Vanilla Memory used: 135.9921875 MB
===== (8, 1024, 16) =====
Optimized Memory used: 5.03125 MB
Vanilla Memory used: 132.0 MB
===== (8, 1024, 32) =====
Optimized Memory used: 10.03125 MB
Vanilla Memory used: 136.0 MB
===== (32, 127, 16) =====
Optimized Memory used: 2.49609375 MB
Vanilla Memory used: 9.861328125 MB
===== (32, 127, 32) =====
Optimized Memory used: 4.9765625 MB
Vanilla Memory used: 11.845703125 MB
===== (32, 128, 16) =====
Optimized Memory used: 2.515625 MB
Vanilla Memory used: 10.0 MB
===== (32, 128, 32) =====
Optimized Memory used: 5.015625 MB
Vanilla Memory used: 12.0 MB
===== (32, 512, 16) =====
Optimized Memory used: 10.0625 MB
Vanilla Memory used: 136.0 MB
===== (32, 512, 32) =====
Optimized Memory used: 20.0625 MB
Vanilla Memory used: 144.0 MB
===== (32, 513, 16) =====
Optimized Memory used: 13.9453125 MB
Vanilla Memory used: 140.2548828125 MB
===== (32, 513, 32) =====
Optimized Memory used: 23.08056640625 MB
Vanilla Memory used: 147.51171875 MB
===== (32, 1023, 16) =====
Optimized Memory used: 21.130859375 MB
Vanilla Memory used: 529.001953125 MB
===== (32, 1023, 32) =====
Optimized Memory used: 40.11328125 MB
Vanilla Memory used: 543.9765625 MB
===== (32, 1024, 16) =====
Optimized Memory used: 21.138671875 MB
Vanilla Memory used: 529.013671875 MB
===== (32, 1024, 32) =====
Optimized Memory used: 40.125 MB
Vanilla Memory used: 544.0 MB
===== (256, 127, 16) =====
Optimized Memory used: 20.0615234375 MB
Vanilla Memory used: 79.470703125 MB
===== (256, 127, 32) =====
Optimized Memory used: 40.216796875 MB
Vanilla Memory used: 95.6572265625 MB
===== (256, 128, 16) =====
Optimized Memory used: 20.234375 MB
Vanilla Memory used: 80.109375 MB
===== (256, 128, 32) =====
Optimized Memory used: 40.125 MB
Vanilla Memory used: 96.0 MB
===== (256, 512, 16) =====
Optimized Memory used: 80.5 MB
Vanilla Memory used: 1088.0 MB
===== (256, 512, 32) =====
Optimized Memory used: 160.5 MB
Vanilla Memory used: 1152.0 MB
===== (256, 513, 16) =====
Optimized Memory used: 80.6572265625 MB
Vanilla Memory used: 1096.125 MB
===== (256, 513, 32) =====
Optimized Memory used: 160.8134765625 MB
Vanilla Memory used: 1160.25 MB
===== (256, 1023, 16) =====
Optimized Memory used: 160.9990234375 MB
Vanilla Memory used: 4216.03515625 MB
===== (256, 1023, 32) =====
Optimized Memory used: 320.8427734375 MB
Vanilla Memory used: 4343.91015625 MB
===== (256, 1024, 16) =====
Optimized Memory used: 161.0 MB
Vanilla Memory used: 4224.0 MB
===== (256, 1024, 32) =====
Optimized Memory used: 321.0 MB
Vanilla Memory used: 4352.0 MB

[------------------- attention -------------------]
                           |  optimized  |  vanilla
1 threads: ----------------------------------------
      B=1, M=127, K=16     |     183.7   |    237.8
      B=1, M=127, K=32     |     183.9   |    239.5
      B=1, M=128, K=16     |     146.0   |    168.1
      B=1, M=128, K=32     |     144.5   |    167.1
      B=1, M=512, K=16     |     142.1   |    165.5
      B=1, M=512, K=32     |     142.1   |    165.0
      B=1, M=513, K=16     |     146.7   |    171.5
      B=1, M=513, K=32     |     146.8   |    171.6
      B=1, M=1023, K=16    |     156.5   |    186.3
      B=1, M=1023, K=32    |     229.5   |    211.8
      B=1, M=1024, K=16    |     166.0   |    195.8
      B=1, M=1024, K=32    |     208.0   |    184.9
      B=8, M=127, K=16     |     143.0   |    165.8
      B=8, M=127, K=32     |     141.2   |    165.9
      B=8, M=128, K=16     |     141.2   |    174.7
      B=8, M=128, K=32     |     149.3   |    206.2
      B=8, M=512, K=16     |     223.7   |    399.3
      B=8, M=512, K=32     |     380.1   |    408.9
      B=8, M=513, K=16     |     331.0   |    419.5
      B=8, M=513, K=32     |     527.2   |    432.3
      B=8, M=1023, K=16    |    1018.1   |   1220.1
      B=8, M=1023, K=32    |    1552.4   |   1243.4
      B=8, M=1024, K=16    |     752.7   |   1206.7
      B=8, M=1024, K=32    |    1324.3   |   1234.4
      B=32, M=127, K=16    |     142.0   |    167.4
      B=32, M=127, K=32    |     141.5   |    169.9
      B=32, M=128, K=16    |     146.4   |    168.7
      B=32, M=128, K=32    |     141.8   |    165.5
      B=32, M=512, K=16    |     766.8   |   1180.1
      B=32, M=512, K=32    |    1362.6   |   1227.0
      B=32, M=513, K=16    |    1201.3   |   1288.6
      B=32, M=513, K=32    |    1950.0   |   1359.9
      B=32, M=1023, K=16   |    3996.6   |   4392.1
      B=32, M=1023, K=32   |    5977.2   |   4546.1
      B=32, M=1024, K=16   |    2939.9   |   4347.4
      B=32, M=1024, K=32   |    5110.6   |   4508.7
      B=256, M=127, K=16   |     581.3   |    629.5
      B=256, M=127, K=32   |     946.7   |    706.3
      B=256, M=128, K=16   |     446.4   |    622.7
      B=256, M=128, K=32   |     828.6   |    695.7
      B=256, M=512, K=16   |    6017.2   |   8183.5
      B=256, M=512, K=32   |   10432.4   |   8744.9
      B=256, M=513, K=16   |    9561.7   |   9208.7
      B=256, M=513, K=32   |   15071.8   |   9893.5
      B=256, M=1023, K=16  |   31757.3   |  32366.8
      B=256, M=1023, K=32  |   47725.4   |  33994.0
      B=256, M=1024, K=16  |   23353.4   |  32033.4
      B=256, M=1024, K=32  |   40533.0   |  33811.4

Times are in microseconds (us).
```

</details>

Fixes https://github.com/facebookresearch/xformers/issues/161.

### DIFF
--- a/tests/test_mem_eff_attention.py
+++ b/tests/test_mem_eff_attention.py
@@ -49,3 +49,58 @@ def test_key_query_all_ones(device, q_len, kv_len, batch_size, k_len):
     ref = value.mean(1, keepdim=True).expand_as(query)
 
     assert torch.allclose(out, ref, atol=1e-5)
+
+
+@pytest.mark.parametrize("k_len", [5, 6, 32])
+@pytest.mark.parametrize("batch_size", [1, 4])
+@pytest.mark.parametrize("kv_len", [3, 15, 32, 33])
+@pytest.mark.parametrize("q_len", [2, 3, 5])
+@pytest.mark.parametrize("device", _devices)
+def test_logsumexp(device, q_len, kv_len, batch_size, k_len):
+    scale = 3
+    query = torch.randn((batch_size, q_len, k_len), device=device) * scale
+    key = torch.randn((batch_size, kv_len, k_len), device=device) * scale
+    value = torch.randn((batch_size, kv_len, k_len), device=device) * scale
+
+    _, lse = torch.ops.xformers.efficient_attention(query, key, value, True)
+    ref_lse = ((query / k_len ** 0.5) @ key.transpose(-2, -1)).logsumexp(-1)
+
+    assert torch.allclose(lse, ref_lse, atol=2e-4)
+
+
+@pytest.mark.parametrize("k_len", [5, 6, 32])
+@pytest.mark.parametrize("batch_size", [1, 4])
+@pytest.mark.parametrize("kv_len", [3, 15, 32, 33])
+@pytest.mark.parametrize("q_len", [2, 3, 5])
+@pytest.mark.parametrize("device", _devices)
+def test_memory_efficient_attention_backward(device, q_len, kv_len, batch_size, k_len):
+    scale = 3
+    query = torch.randn((batch_size, q_len, k_len), device=device) * scale
+    key = torch.randn((batch_size, kv_len, k_len), device=device) * scale
+    value = torch.randn((batch_size, kv_len, k_len), device=device) * scale
+
+    query.requires_grad_(True)
+    key.requires_grad_(True)
+    value.requires_grad_(True)
+
+    out = xformers.ops.memory_efficient_attention(query, key, value)
+    out.backward(torch.ones_like(query))
+
+    grad_q = query.grad
+    grad_k = key.grad
+    grad_v = value.grad
+
+    query.grad = None
+    key.grad = None
+    value.grad = None
+
+    ref = ref_attention(query, key, value)
+    ref.backward(torch.ones_like(query))
+
+    # there is some extra precision loss in the CPU implementation due to an
+    # extra accumulation step in grad_q, which is not present in the CUDA
+    # implementation
+    atol = 3e-4 if device == "cuda" else 4e-4
+    assert torch.allclose(grad_q, query.grad, atol=atol), "grad_q doesn't match"
+    assert torch.allclose(grad_k, key.grad, atol=atol), "grad_k doesn't match"
+    assert torch.allclose(grad_v, value.grad, atol=atol), "grad_v doesn't match"
--- a/xformers/benchmarks/benchmark_mem_eff_attention.py
+++ b/xformers/benchmarks/benchmark_mem_eff_attention.py
@@ -26,6 +26,7 @@ NUM_THREADS = [1] if device.type == "cuda" else [1, 40]
 SHAPES = list(
     itertools.product([1, 8, 32, 256], [127, 128, 512, 513, 1023, 1024], [16, 32])
 )
+SHAPES = [(256, 1024, 32)]
 
 results = []
 mem_use: Dict[str, Dict[str, float]] = dict(optimized={}, vanilla={})
@@ -38,7 +39,7 @@ for num_threads in NUM_THREADS:
         q = torch.rand(shape, device=device)
         sub_label = f"B={B}, M={M}, K={K}"
 
-        if True:
+        if False:
             r = xformers.ops.memory_efficient_attention(q, q, q)
 
             rr = ref_attention(q, q, q)
@@ -51,7 +52,7 @@ for num_threads in NUM_THREADS:
                 stmt="fn(q, q, q)",
                 globals={
                     "q": q,
-                    "fn": torch.ops.xformers.efficient_attention,
+                    "fn": xformers.ops.memory_efficient_attention,
                 },
                 label="attention",
                 description="optimized",
--- a/xformers/benchmarks/benchmark_mem_eff_attention.py
+++ b/xformers/benchmarks/benchmark_mem_eff_attention.py
@@ -26,71 +26,151 @@ NUM_THREADS = [1] if device.type == "cuda" else [1, 40]
 SHAPES = list(
     itertools.product([1, 8, 32, 256], [127, 128, 512, 513, 1023, 1024], [16, 32])
 )
-SHAPES = [(256, 1024, 32)]
 
 results = []
 mem_use: Dict[str, Dict[str, float]] = dict(optimized={}, vanilla={})
 
-print(f"Processing {len(SHAPES)} cases")
-for num_threads in NUM_THREADS:
-    for shape in SHAPES:
-        print(f"===== {shape} =====")
-        B, M, K = shape
-        q = torch.rand(shape, device=device)
-        sub_label = f"B={B}, M={M}, K={K}"
 
-        if False:
-            r = xformers.ops.memory_efficient_attention(q, q, q)
+def benchmark_forward():
+    print(f"Processing {len(SHAPES)} cases")
+    print("Forward")
+    for num_threads in NUM_THREADS:
+        for shape in SHAPES:
+            print(f"===== {shape} =====")
+            B, M, K = shape
+            q = torch.rand(shape, device=device)
+            sub_label = f"B={B}, M={M}, K={K}"
 
-            rr = ref_attention(q, q, q)
-            assert (r - rr).abs().max() < 1e-5
+            if True:
+                r = xformers.ops.memory_efficient_attention(q, q, q)
 
-        torch.cuda.reset_peak_memory_stats()
-        torch.cuda.synchronize()
-        results.append(
-            benchmark.Timer(
-                stmt="fn(q, q, q)",
-                globals={
-                    "q": q,
-                    "fn": xformers.ops.memory_efficient_attention,
-                },
-                label="attention",
-                description="optimized",
-                sub_label=sub_label,
-                num_threads=num_threads,
-            ).blocked_autorange(min_run_time=min_run_time)
-        )
-        torch.cuda.synchronize()
-        memory = torch.cuda.max_memory_allocated() / 2 ** 20
-        mem_use["optimized"][sub_label] = memory
-        memory_str = f"Memory used: {memory} MB"
+                rr = ref_attention(q, q, q)
+                assert (r - rr).abs().max() < 1e-5
 
-        print("Optimized", memory_str)
+            torch.cuda.reset_peak_memory_stats()
+            torch.cuda.synchronize()
+            results.append(
+                benchmark.Timer(
+                    stmt="fn(q, q, q)",
+                    globals={
+                        "q": q,
+                        "fn": xformers.ops.memory_efficient_attention,
+                    },
+                    label="attention",
+                    description="optimized",
+                    sub_label=sub_label,
+                    num_threads=num_threads,
+                ).blocked_autorange(min_run_time=min_run_time)
+            )
+            torch.cuda.synchronize()
+            memory = torch.cuda.max_memory_allocated() / 2 ** 20
+            mem_use["optimized"][sub_label] = memory
+            memory_str = f"Memory used: {memory} MB"
 
-        torch.cuda.reset_peak_memory_stats()
-        torch.cuda.synchronize()
-        results.append(
-            benchmark.Timer(
-                stmt="fn(q, q, q)",
-                globals={
-                    "q": q,
-                    "fn": ref_attention,
-                },
-                label="attention",
-                description="vanilla",
-                sub_label=sub_label,
-                num_threads=num_threads,
-            ).blocked_autorange(min_run_time=min_run_time)
-        )
+            print("Optimized", memory_str)
 
-        torch.cuda.synchronize()
-        memory = torch.cuda.max_memory_allocated() / 2 ** 20
-        mem_use["vanilla"][sub_label] = memory
-        memory_str = f"Memory used: {memory} MB"
-        print("Vanilla", memory_str)
+            torch.cuda.reset_peak_memory_stats()
+            torch.cuda.synchronize()
+            results.append(
+                benchmark.Timer(
+                    stmt="fn(q, q, q)",
+                    globals={
+                        "q": q,
+                        "fn": ref_attention,
+                    },
+                    label="attention",
+                    description="vanilla",
+                    sub_label=sub_label,
+                    num_threads=num_threads,
+                ).blocked_autorange(min_run_time=min_run_time)
+            )
+
+            torch.cuda.synchronize()
+            memory = torch.cuda.max_memory_allocated() / 2 ** 20
+            mem_use["vanilla"][sub_label] = memory
+            memory_str = f"Memory used: {memory} MB"
+            print("Vanilla", memory_str)
+
+    compare = benchmark.Compare(results)
+    compare.print()
+
+    pprint.pprint(mem_use)
 
 
-compare = benchmark.Compare(results)
-compare.print()
+def benchmark_backward():
+    print(f"Processing {len(SHAPES)} cases")
+    print("Backward")
+    for num_threads in NUM_THREADS:
+        for shape in SHAPES:
+            print(f"===== {shape} =====")
+            B, M, K = shape
+            q = torch.rand(shape, device=device, requires_grad=True)
+            sub_label = f"B={B}, M={M}, K={K}"
 
-pprint.pprint(mem_use)
+            if True:
+                r = xformers.ops.memory_efficient_attention(q, q, q)
+                r.backward(torch.ones_like(q))
+
+                grad = q.grad
+                q.grad = None
+
+                rr = ref_attention(q, q, q)
+                rr.backward(torch.ones_like(q))
+                assert (grad - q.grad).abs().max() < 1e-5
+
+            out = xformers.ops.memory_efficient_attention(q, q, q)
+            grad = torch.ones_like(q)
+
+            torch.cuda.reset_peak_memory_stats()
+            torch.cuda.synchronize()
+            results.append(
+                benchmark.Timer(
+                    stmt="out.backward(grad, retain_graph=True)",
+                    globals={
+                        "out": out,
+                        "grad": grad,
+                    },
+                    label="attention",
+                    description="optimized",
+                    sub_label=sub_label,
+                    num_threads=num_threads,
+                ).blocked_autorange(min_run_time=min_run_time)
+            )
+            torch.cuda.synchronize()
+            memory = torch.cuda.max_memory_allocated() / 2 ** 20
+            mem_use["optimized"][sub_label] = memory
+            memory_str = f"Memory used: {memory} MB"
+
+            print("Optimized", memory_str)
+
+            out = ref_attention(q, q, q)
+            torch.cuda.reset_peak_memory_stats()
+            torch.cuda.synchronize()
+            results.append(
+                benchmark.Timer(
+                    stmt="out.backward(grad, retain_graph=True)",
+                    globals={
+                        "out": out,
+                        "grad": grad,
+                    },
+                    label="attention",
+                    description="vanilla",
+                    sub_label=sub_label,
+                    num_threads=num_threads,
+                ).blocked_autorange(min_run_time=min_run_time)
+            )
+
+            torch.cuda.synchronize()
+            memory = torch.cuda.max_memory_allocated() / 2 ** 20
+            mem_use["vanilla"][sub_label] = memory
+            memory_str = f"Memory used: {memory} MB"
+            print("Vanilla", memory_str)
+
+    compare = benchmark.Compare(results)
+    compare.print()
+
+    pprint.pprint(mem_use)
+
+
+benchmark_forward()
+benchmark_backward()
--- a/xformers/components/attention/csrc/cpu/attention.cpp
+++ b/xformers/components/attention/csrc/cpu/attention.cpp
@@ -147,10 +147,189 @@ at::Tensor attention(
   return res;
 }
 
+template <typename scalar_t>
+void attention_backward_kernel(
+    at::TensorAccessor<scalar_t, 3> grad_q,
+    at::TensorAccessor<scalar_t, 3> grad_k,
+    at::TensorAccessor<scalar_t, 3> grad_v,
+    at::TensorAccessor<scalar_t, 3> grad_out,
+    at::TensorAccessor<scalar_t, 3> q,
+    at::TensorAccessor<scalar_t, 3> k,
+    at::TensorAccessor<scalar_t, 3> v,
+    at::TensorAccessor<scalar_t, 2> logsumexp_normalizer,
+    at::TensorAccessor<scalar_t, 3> buffer,
+    at::TensorAccessor<scalar_t, 3> buffer2,
+    at::TensorAccessor<scalar_t, 3> buffer3 //,
+    // at::TensorAccessor<int64_t, 2> mask
+) {
+  int64_t K = q.size(2);
+  int64_t B = q.size(0);
+  int64_t M = q.size(1);
+  int64_t N = k.size(1);
+  int64_t grain_size = 1; // buffer.size(1);
+  at::parallel_for(0, B, grain_size, [&](int64_t start, int64_t end) {
+    auto buf = buffer[at::get_thread_num()][0];
+    auto buf2 = buffer2[at::get_thread_num()];
+    auto buf3 = buffer3[at::get_thread_num()][0];
+    for (int64_t i = start; i < end; i++) {
+      for (int64_t l = 0; l < N; l++) {
+        for (int64_t k = 0; k < K; k++) {
+          buf2[l][k] = 0;
+        }
+      }
+
+      for (int64_t j = 0; j < M; j++) {
+        for (int64_t k = 0; k < K; k++) {
+          buf[k] = 0;
+        }
+        auto aar = q[i][j];
+        auto normalizer = logsumexp_normalizer[i][j];
+        scalar_t tmp_sum = 0;
+        for (int64_t l = 0; l < N; l++) {
+          auto bar = k[i][l];
+          scalar_t si = 0;
+          for (int64_t k = 0; k < K; k++) {
+            si += aar[k] * bar[k];
+          }
+          scalar_t attn_v = std::exp(si - normalizer);
+
+          for (int64_t k = 0; k < K; k++) {
+            grad_v[i][l][k] += attn_v * grad_out[i][j][k];
+          }
+
+          // now compute grad_q and grad_k
+          scalar_t g_attn = 0;
+          for (int64_t k = 0; k < K; k++) {
+            g_attn += grad_out[i][j][k] * v[i][l][k];
+            // g_attn[i][j][l] += grad_out[i][j][k] * v[i][l][k];
+          }
+          scalar_t tmp = attn_v * g_attn;
+          tmp_sum += tmp;
+
+          // grad_q is easy
+          for (int64_t k = 0; k < K; k++) {
+            grad_q[i][j][k] += tmp * bar[k]; // bar == key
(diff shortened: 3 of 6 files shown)


blefaudeux:
it's great, thanks @fmassa ! Open question: how would it make the more sense to integrate it with the rest, for people who build from the registers ? I can add a follow up PR to add that to the existing attention mechanisms, or it could just be a flag for "scaled dot product" (or something else ?)

tests/test_mem_eff_attention.py
```diff
@@ -49,3 +49,58 @@ def test_key_query_all_ones(device, q_len, kv_len, batch_size, k_len):
...
+    grad_v = value.grad
+
+    query.grad = None
+    key.grad = None
+    value.grad = None
+
+    ref = ref_attention(query, key, value)
+    ref.backward(torch.ones_like(query))
+
+    # there is some extra precision loss in the CPU implementation due to an
```
blefaudeux, line 100: thanks for the comment, helpful !

xformers/benchmarks/benchmark_mem_eff_attention.py
```diff
@@ -30,66 +30,147 @@ def ref_attention(q, k, v):
...
+    print(f"Processing {len(SHAPES)} cases")
+    print("Forward")
+    for num_threads in NUM_THREADS:
+        for shape in SHAPES:
+            print(f"===== {shape} =====")
+            B, M, K = shape
+            q = torch.rand(shape, device=device)
+            sub_label = f"B={B}, M={M}, K={K}"
+
+            if True:
```
blefaudeux, line 44: debug ?

xformers/benchmarks/benchmark_mem_eff_attention.py
```diff
@@ -30,66 +30,147 @@ def ref_attention(q, k, v):
...
+            memory_str = f"Memory used: {memory} MB"
+            print("Vanilla", memory_str)
+
+    compare = benchmark.Compare(results)
+    compare.print()
+
+    pprint.pprint(mem_use)
+
+
+benchmark_forward()
```
blefaudeux, line 175: [nit] maybe possible to factorize the two, but not super important, good tool to have already !

xformers/components/attention/csrc/cpu/attention.cpp
```diff
@@ -28,10 +28,12 @@ scalar_t max(scalar_t* buf) {
...
 void attention_kernel(
     at::TensorAccessor<scalar_t, 3> output,
+    at::TensorAccessor<scalar_t, 2> logsumexp,
     at::TensorAccessor<scalar_t, 3> query,
     at::TensorAccessor<scalar_t, 3> key,
     at::TensorAccessor<scalar_t, 3> value,
-    at::TensorAccessor<scalar_t, 3> buffer //,
+    at::TensorAccessor<scalar_t, 3> buffer,
+    bool compute_logsumexp
     // at::TensorAccessor<int64_t, 2> mask
```
blefaudeux, line 37: ahah, next step is to make it sparse ? :D (not that much of a joke..)

xformers/components/attention/csrc/cpu/attention.cpp
```diff
@@ -90,15 +92,18 @@ void attention_kernel(
         for (int64_t k = 0; k < K; k++) {
           oo[k] = buf[k] / s_prime;
         }
+        if (compute_logsumexp)
+          logsumexp[i][j] = m_prime + std::log(s_prime);
```
blefaudeux, line 96: ah, I didn't think of that, this is why your backward is so fast, and it does not weight that much actually (one per line, not the whole attention map)

xformers/components/attention/csrc/cuda/attention.cu
```diff
@@ -58,6 +59,25 @@ __device__ __forceinline__ void iDiv(scalar_t x1, float* out) {
   out[0] /= x1;
 }
 
+template <typename scalar_t>
+__device__ __forceinline__ void myGpuAtomicAdd(scalar_t* address, float4 val) {
```
blefaudeux, line 63: ah interesting, I thought that this was a little costly and not too much of a good idea in practice, looks like it's not a good intuition.. We're doing this (accumulate across threads) on the triton side for layernorm, but could be a good idea to extend this for linear layer / bias gradient, I'll have a look when I get the time

xformers/components/attention/csrc/cuda/attention.cu
```diff
@@ -393,6 +416,7 @@ __global__ void attention_kernel(
     output_block[q_item_idx] =
         reinterpret_cast<vec_t*>(output[batch_idx][index].data());
     m_prime[q_item_idx] = -std::numeric_limits<scalar_t>::infinity();
+    logsumexp_block[q_item_idx] = &logsumexp[batch_idx][index];
```
blefaudeux, line 419: I know it was already there in the FW PR, but I missed this, could you elaborate ? You're collecting all the pointers as a first step ? Necessary for the pragma unroll down the line ? Trying to understand this cuda trick :)

xformers/components/attention/csrc/cuda/attention.cu
```diff
@@ -473,56 +497,30 @@ __global__ void attention_kernel(
       output_block[q_item_idx][k] = tmp;
     }
   }
+
+  if (compute_logsumexp) {
+#pragma unroll
```
blefaudeux, line 502: guess is that the pointer array is because of this ?

xformers/components/attention/csrc/cuda/attention.cu
```diff
@@ -577,14 +579,611 @@ at::Tensor attention(
...
+  int64_t B = query.size(0);
+  int64_t M = query.size(1);
+  int64_t N = key.size(1);
+  int64_t K = query.size(2);
+
+  at::Tensor res = at::zeros({B, M, K}, query.options());
+  at::Tensor logsumexp = at::empty({B, M}, query.options());
+
+  // have to pass compute_logsumexp as a template parameter
+  // otherwise there is a slowdown in the kernel...
```
blefaudeux, line 641: register spilling ? (like, it inlines compute  logsumexp + a branch, and it takes too much space ?)

xformers/components/attention/csrc/cuda/attention.cu
```diff
@@ -577,14 +579,611 @@ at::Tensor attention(
...
+
+  for (int64_t k = 0; k < K / kVecSize; k += 1) {
+#pragma unroll
+    for (int k_item_idx = 0; k_item_idx < kBlockSizeK; k_item_idx++) {
+      vec_t kk = __ldg(kb[k_item_idx] + k);
+      iMul(scale, &kk);
+      vec_t tt = __ldg(vb[k_item_idx] + k);
+#pragma unroll
+      for (int q_item_idx = 0; q_item_idx < kBlockSizeQ; q_item_idx++) {
+        sputnik::VectorCompute<vec_t>::Dot(
```
blefaudeux, line 744: (I guess that `computeDot` is the most interesting place for that, but I cannot comment there.) It could be interesting to pull in tensor cores for the .dot() for fp16, and for fp32 on newer hardware (A100 +), I don't know if there are existing primitives to preferably use there (from Cuda/cudnn/cublas, [like in this example](https://developer.nvidia.com/blog/programming-tensor-cores-cuda-9/) or from torch). 

Maybe that it's not needed actually, sorry for bringing these tensor cores up all the time but just trying to think ahead for the possible next steps (and possibly hidden caveats if you developed on a P100)

xformers/benchmarks/benchmark_mem_eff_attention.py
```diff
@@ -30,66 +30,147 @@ def ref_attention(q, k, v):
...
+            print(f"===== {shape} =====")
+            B, M, K = shape
+            q = torch.rand(shape, device=device)
+            sub_label = f"B={B}, M={M}, K={K}"
+
+            if True:
+                r = xformers.ops.memory_efficient_attention(q, q, q)
+
+                rr = ref_attention(q, q, q)
+                assert (r - rr).abs().max() < 1e-5
```
blefaudeux, line 48: this does not pass on a 3080 / cuda 11.6, could be interesting to test with the T4s on CircleCi, could well be because of TF32 (you would need to switch the torch flag forcing fp32 computations). Implicitly this probably means that the torch implementation switched to tensor cores I think, which changes the time difference in between the two implementations (but not a fundamental issue)

xformers/benchmarks/benchmark_mem_eff_attention.py
```diff
@@ -30,66 +30,147 @@ def ref_attention(q, k, v):
...
+            if True:
+                r = xformers.ops.memory_efficient_attention(q, q, q)
+                r.backward(torch.ones_like(q))
+
+                grad = q.grad
+                q.grad = None
+
+                rr = ref_attention(q, q, q)
+                rr.backward(torch.ones_like(q))
+                assert (grad - q.grad).abs().max() < 1e-5
```
blefaudeux, line 119: same as above, this does not pass on a 3080, guess is because of tf32 vs. float32 (would be the)

xformers/components/attention/csrc/cuda/attention.cu
```diff
@@ -577,14 +579,611 @@ at::Tensor attention(
...
+
+template <
+    typename scalar_t,
+    typename vec_t,
+    int kBlockSizeQ,
+    int kBlockSizeK,
+    int TILE_SIZEQ,
+    int TILE_SIZEK,
+    bool check_bounds>
+__global__ void attention_backward_grad_v_kernel(
```
blefaudeux, line 661: a bit harder for me to follow since it's not something that I dived into (vs. the fw pass) so I'm forced to skim a liltte.. looks clean as always

xformers/components/attention/csrc/cuda/attention.cu
```diff
@@ -577,14 +579,611 @@ at::Tensor attention(
...
+  int64_t l = blockIdx.y * blockDim.y * kBlockSizeK + threadIdx.y * kBlockSizeK;
+
+  __shared__ scalar_t fact[TILE_SIZEQ][TILE_SIZEK + 1];
+
+#pragma unroll
+  for (int k_item_idx = 0; k_item_idx < kBlockSizeK; k_item_idx++) {
+#pragma unroll
+    for (int q_item_idx = 0; q_item_idx < kBlockSizeQ; q_item_idx++) {
+      fact[kBlockSizeQ * threadIdx.x + q_item_idx]
+          [kBlockSizeK * threadIdx.y + k_item_idx] = 0;
```
blefaudeux, line 688: no faster call than that to zero the buffer ?

xformers/components/attention/csrc/cuda/attention.cu
```diff
@@ -577,14 +579,611 @@ at::Tensor attention(
...
+    }
+  }
+
+  scalar_t normalizer[kBlockSizeQ];
+  scalar_t tmp_sum[kBlockSizeQ] = {0};
+
+  vec_t *qb[kBlockSizeQ], *kb[kBlockSizeK], *vb[kBlockSizeK], *gb[kBlockSizeQ],
```
blefaudeux, line 695: same as the other one, collect all pointers first then proceed, ok. It's typical I

blefaudeux (approved):
looks massively great to me, thanks @fmassa ! Couple of questions, mostly to understand better on my side and ask for possible follow ups down the line

blefaudeux:
if that helps, speed report on a 3080 

`[------------------- attention -------------------]
                           |  optimized  |  vanilla
1 threads: ----------------------------------------
      B=1, M=127, K=16     |      19.9   |     54.1
      B=1, M=127, K=32     |      25.2   |     49.5
      B=1, M=128, K=16     |      19.3   |     52.2
      B=1, M=128, K=32     |      19.9   |     50.1
      B=1, M=512, K=16     |      35.3   |     55.2
      B=1, M=512, K=32     |      63.7   |     63.0
      B=1, M=513, K=16     |      35.9   |     58.6
      B=1, M=513, K=32     |      64.4   |     63.7
      B=1, M=1023, K=16    |      69.6   |     55.1
      B=1, M=1023, K=32    |     127.3   |     59.6
      B=1, M=1024, K=16    |      66.0   |     58.8
      B=1, M=1024, K=32    |     121.7   |     65.7
      B=8, M=127, K=16     |      18.7   |     55.3
      B=8, M=127, K=32     |      25.4   |     49.9
      B=8, M=128, K=16     |      21.5   |     58.7
      B=8, M=128, K=32     |      20.1   |     59.3
      B=8, M=512, K=16     |      66.4   |    116.1
      B=8, M=512, K=32     |     126.5   |    121.6
      B=8, M=513, K=16     |      67.5   |    132.5
      B=8, M=513, K=32     |     130.3   |    145.7
      B=8, M=1023, K=16    |     256.2   |    427.2
      B=8, M=1023, K=32    |     537.2   |    452.1
      B=8, M=1024, K=16    |     249.7   |    397.4
      B=8, M=1024, K=32    |     518.8   |    410.4
      B=32, M=127, K=16    |      25.8   |     55.8
      B=32, M=127, K=32    |      48.9   |     50.7
      B=32, M=128, K=16    |      22.1   |     59.6
      B=32, M=128, K=32    |      39.5   |     59.2
      B=32, M=512, K=16    |     259.7   |    406.6
      B=32, M=512, K=32    |     534.8   |    425.6
      B=32, M=513, K=16    |     263.9   |    464.9
      B=32, M=513, K=32    |     544.2   |    510.8
      B=32, M=1023, K=16   |     948.7   |   1629.7
      B=32, M=1023, K=32   |    2027.9   |   1729.1
      B=32, M=1024, K=16   |     916.9   |   1510.2
      B=32, M=1024, K=32   |    1945.0   |   1573.7
      B=256, M=127, K=16   |     238.6   |    261.0
      B=256, M=127, K=32   |     403.1   |    339.7
      B=256, M=128, K=16   |     165.9   |    236.1
      B=256, M=128, K=32   |     316.9   |    261.1
      B=256, M=512, K=16   |    1870.8   |   3078.6
      B=256, M=512, K=32   |    3908.3   |   3236.4
      B=256, M=513, K=16   |    1927.2   |   3638.1
      B=256, M=513, K=32   |    4016.3   |   3922.3
      B=256, M=1023, K=16  |    7383.7   |  12697.6
      B=256, M=1023, K=32  |   15664.0   |  13622.2
      B=256, M=1024, K=16  |    7172.9   |  11887.2
      B=256, M=1024, K=32  |   15038.3   |  12385.9

Times are in microseconds (us).
`

blefaudeux:
Another follow up question is about the supported K, I think that 256 is typical in NLP / big models for instance, or 64 / 128 for the GPT series. Certainly not blocking, but it's a request bound to come up :)

xformers/benchmarks/benchmark_mem_eff_attention.py
fmassa: Yeah, it's a debug flag that is sometimes helpful: sometimes I "break" the kernel by removing some parts of the computation and see what speedup I would get. But doing so means that the computation won't be correct anymore, so it was useful to just disable correctness checks.

I can remove this in if you want, but as I expect to still do some more performance tuning, I'd like to keep this around for a bit longer if it's ok with you?

xformers/benchmarks/benchmark_mem_eff_attention.py
fmassa: yeah, it's totally possible. I've also added in a separate branch a `benchmark_forward_and_backward` case, and it started to have quite a bit of duplication. I can look into refactoring this up in a follow-up PR

xformers/components/attention/csrc/cpu/attention.cpp
fmassa: yes, integrating sparsity is in the plans :-) But before, I'll look into the `K > 32` case

xformers/components/attention/csrc/cpu/attention.cpp
fmassa: yeah, keeping this temporary around made the backward kernel faster and also easier to implement, so I decided that it was worth the extra memory.

Also, in a follow-up PR I'll avoid allocating the `logsumexp` buffer in the function if `compute_logsumexp` is `false`, so that this is only a memory price to pay during training

xformers/components/attention/csrc/cuda/attention.cu
fmassa: I was also a bit afraid of the atomicAdds in the beginning, but it turned out to not be too slow. Plus it made it easier to parallelize the kernel, so why not.

xformers/components/attention/csrc/cuda/attention.cu
fmassa: The main benefit of this preamble is to handle sequence lengths which are not multiple of 32 (or the block size which I'm using). Instead of handling the out-of-bonds directly in the hotpaths of the kernel, I handle it beforehand and repeat the last element if needed, so I don't index out of the bounds of the kernel.
There are probably other / better ways of handling generic sequence lengths, but this was the easiest to implement so I decided to go for it.

xformers/components/attention/csrc/cuda/attention.cu
fmassa: I think it might also help as well, yes, although it was not directly the main reason why I did it

xformers/components/attention/csrc/cuda/attention.cu
fmassa: might be. Maybe I should try this out as a non-inlined function call to see if it make things better. Lots of improvements to be done in the future! :-)

xformers/components/attention/csrc/cuda/attention.cu
fmassa: TensorCores are going to be a very important next step indeed, and I'm already looking into how to use them. But before that I think it might be better to support the `K > 32` case first, wdyt?

xformers/benchmarks/benchmark_mem_eff_attention.py
fmassa: Good point, I should probably change those defaults, or just disable `TF32` in the benchmarks (but that makes it for slower baselines), or just disable this correctness check by default. Which one would you prefer?

xformers/components/attention/csrc/cuda/attention.cu
fmassa: This is a good point, I should probably write some comments in the top of the functions with the implementation that this is actually doing, as a PyTorch code for ease of read. If you don't mind, I'd like to get this PR merged now, but I can add more comments with the `K > 32` PR that I'm working on

xformers/components/attention/csrc/cuda/attention.cu
fmassa: You mean with `memset`, or with default initializer (like `= {};`) ?

It turns out that shared memory doesn't allow for default initializers, so I let all threads in the block to collaboratively zero the buffer. From some basic profiling, this part of the code doesn't seem to take any noticeable time, so I decided to leave it like that.

Also worth noting that I need to zero this only in the cases where the input sequence length is not a multiple of 32, otherwise there would be uninitialized values in there. Now that we have a template parameter in the kernel for this case, I could probably also put this under an `if (check_bounds)`, but I wouldn't expect it to bring noticeable improvements.

xformers/components/attention/csrc/cuda/attention.cu
fmassa: Yeah, it was to simplify a bit the handling of the code in other parts of the kernel. Maybe it saves on a couple of instructions, so it might be slightly faster to do this.

fmassa:
Thanks for the timely review @blefaudeux !

Yes, the `K > 32` case is definitely super important and I'm looking into how to address this. Naively bumping up `BUFFER_SIZE` is not ideal, and our kernels end up being 2x slower than the baseline, so there is definitely some room for improvement here.

Using TensorCores is also next in my optimization list, which I hope to get to after I'm back from holidays.

xformers/benchmarks/benchmark_mem_eff_attention.py
blefaudeux: it's totally ok, flagging it just in case but understood, no worries

xformers/components/attention/csrc/cuda/attention.cu
blefaudeux: ahh I didn't see that at first sight, makes sense. I don't know of a better way myself (except that maybe that CUDA has similar concepts as all the graphics interfaces, where you can clamp memory/textures fetches automatically so that it does not out of bound and repeat or pad). It's been a long time since I wrote in Cuda so by now I just don't know..

xformers/components/attention/csrc/cuda/attention.cu
blefaudeux: agree on bigger K first, at this point the perf impact is not super clear to me (for fp32 at least, on the 3080 which has tensor cores this was very competitive vs. pytorch and tensor cores) while the limitation of a small K is very well defined

xformers/benchmarks/benchmark_mem_eff_attention.py
blefaudeux: I would switch tf32 off here, I think that it's the best correctness check : you assume fp32 in the kernel, let's check correctness against fp32 ? (`torch.backends.cuda.matmul.allow_tf32 = False`)
Good to keep in mind in the benchmarks that the comparison is not iso-accuracy by the way, your implementation is actually more precise :)

xformers/components/attention/csrc/cuda/attention.cu
blefaudeux: done > wild plans, sound good to me !

fmassa:
BTW, I forgot to answer one of your comments:

> how would it make the more sense to integrate it with the rest, for people who build from the registers

My thinking was to enable this directly in `scaled_dot_product_attention`, maybe through a global flag, so that all other attention mechanisms can benefit from it if enabled. I'll do it in a follow-up PR as well (once I get to the other tasks)

blefaudeux:
> BTW, I forgot to answer one of your comments:
> 
> > how would it make the more sense to integrate it with the rest, for people who build from the registers
> 
> My thinking was to enable this directly in `scaled_dot_product_attention`, maybe through a global flag, so that all other attention mechanisms can benefit from it if enabled. I'll do it in a follow-up PR as well (once I get to the other tasks)

sounds good to me, I would have done that also !

PR #282, opened 2022-04-25
Handle non-contiguous grad_out and add checks for qkv

fmassa:
Fixes cases where we do `attention().sum().backward()`, which has a non-contiguous `grad_out`


fmassa:
CPU tests are failing for some reason, I'll have a look tomorrow

qasfb:
lgtm

PR #283, opened 2022-04-26
[perf] Fused linear : small FW cleanup and much better perfs

blefaudeux:
## What does this PR do?
Improves the perfs for the fused linear, mostly the forward pass, but measurable effects all around. The BW pass could be greatly improved, tons of perf on the table, I'll try to work a bit on that in the coming days. This would be applicable for MLP and MHA projection, if we end up writing a dedicated self-attention projection kernel (the op is wx+b, same as fused linear without the activation)

## Before submitting

- [🙃] Did you have fun?

## PR review


blefaudeux:
There's a lot of speed left in the BW, with a fresher pair of eyes it's obvious. This is a very small PR which already brings some decent speed, all from the FW pass. Fist bump @dianaml0, I know you were hoping for more perf there recently. I looked into that while contemplating a MHA - self attention dedicated projection kernel, which would be very similar (without the activation, and dispatching into 3 buffers)

blefaudeux:
in a nutshell.. 

gnome-shell-screenshot-b3jc2b

blefaudeux:
Inference is even nicer (this PR only touches this part)

blefaudeux:
Ideally @dianaml0 the backward pass should be improved, and the FusedMLP could revert to using this instead of the not-super-impactful fused dropout

dianaml0 (approved):
Awesome results!

moinnadeem:
@blefaudeux Hey Ben! I'm curious: did you run this on an end-to-end benchmark? I've been able to recreate the increased FLOPs in a timing measurement, but it performs roughly the same as a PyTorch vanilla linear + GeLU layer on a BERT-Base (110M) on an A100 w/ 80GB VRAM. 

Is this expected? I did install `xformers` from source since the `pip` package doesn't reflect this PR, and use Triton v2 instead of v1.1.

blefaudeux:
https://github.com/facebookresearch/xformers/issues/296 added to make sure that I don't forget to update the curves. Ideally I would love to check in a small perf bump at the same time

blefaudeux:
I'm also beginning to wonder whether this is related to https://github.com/pytorch/pytorch/pull/76509

ngimel:
#76509 landed yesterday, so couldn't affect earlier results. Also, if affects only fp32 matmul ops, not fp16.

blefaudeux:
> #76509 landed yesterday, so couldn't affect earlier results. Also, if affects only fp32 matmul ops, not fp16.

fair enough, I was wondering if in some of the curves I was comparing triton (tf32 internally) vs. pytorch (fp32 at some point, then tf32, now back to fp32). I meant to reference the issue to say that I could have tripped over one of these changes (I was not always tracking pytorch dev, although our CI is)

ngimel:
Pytorch was always tf32 until yesterday it became fp32. But for fp16 inputs it doesn't matter at all, and I see only fp16 curves.

PR #292, opened 2022-05-04
[feat] Natively support PatchEmbedding layers

### LINKED ISSUE #285: Support natively Swin or Poolformer architectures
# 🚀 Feature
These transformers follow a transformation similar to the typical convnets, trading width for latent space the deeper you go. This has proven to be very efficient in the CV domain. It's possible to generate these models right now in xformers but a bit cumbersome, it could be nice to have a small helper for that (adds the pooling and fill in the dimensions for each layer)

## Motivation
Ease of use, make it faster to compare different ideas

## Pitch
Should not be more complicated than ~100 lines (max), pretty orthogonal to everything else (if you're not interested it does not get in the way)

## Alternatives
Not doing it

blefaudeux:
## What does this PR do?
- First step to close #285, we actually need to support the patch embedding part to be able to generate all [Metaformers](https://arxiv.org/pdf/2111.11418.pdf). We could expose more options as of what function is used to "pool" the patches, this PR follows ViT in that it's a Conv2d
- Add a learnable position embedding, somehow not there
- Change the microViT example so that it benefits from this new support
- Testing a "classic" small ViT training on CIFAR, seems alright

TODO:
- [x] Add a matching unit test

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/xformers/factory/block_configs.py
+++ b/xformers/factory/block_configs.py
@@ -1,0 +1,224 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, Dict, Optional
+
+from xformers.components import LayerNormStyle
+from xformers.components.feedforward import FEEDFORWARD_REGISTRY, FeedforwardConfig
+from xformers.components.positional_embedding import (
+    POSITION_EMBEDDING_REGISTRY,
+    PositionEmbeddingConfig,
+)
+from xformers.utils import generate_matching_config
+
+
+class LayerPositionBitmask(int, Enum):
+    First = 0b01
+    Last = 0b10
+    Default = 0b11
+
+
+class LayerPosition:
+    """Bitmask to mark this layer as first, last, nothing or both"""
+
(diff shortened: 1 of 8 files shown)


xformers/factory/block_configs.py
```diff
@@ -0,0 +1,224 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
```
blefaudeux, line 1: 99% of the code is the same as  before, just moved out of "block_factory" which was becoming a little big. A lot of the steps here are pure config machinery, so I figured that block_factory was easier to understand without them

xformers/helpers/hierarchical_configs.py
```diff
@@ -0,0 +1,72 @@
...
+
+@dataclass
+class BasicLayerConfig:
+    embedding: int
+    n_heads: int
+    pooling_ratio: int
+    attention_mechanism: str
+
+
+def get_hierarchical_configuration(
```
blefaudeux, line 21: WIP actually, should not be there probably, see next PR :) (will generate all Metaformers)

xformers/factory/block_configs.py
dianaml0: Nice, feels cleaner!

PR #300, opened 2022-05-10
Fix sddmm2 when nnz=0

fmassa:
One of the internal implementations of sampled dense dense matrix multiplication (sddmm) that we have had two issues:
- It didn't guard kernel launches when nnz=0 (which would yield a grid size of 0 in one of its dimensions)
- it didn't contain a `cudaGetLastError` call in the end of the functions. So errors in this kernel would only be reported at the next function invocation, misleading the true location of the issue.

The PR fixes this by returning early when `nnz=0` (which is fine as there is no data in the tensor anyway), and also add cudaGetLastError which was missing.

Should fix the errors present in https://github.com/facebookresearch/xformers/pull/263

### DIFF
--- a/tests/test_custom_ops.py
+++ b/tests/test_custom_ops.py
@@ -158,14 +158,14 @@ def test_sddmm_sputnik(device):
 
 
 @cuda_only
+@pytest.mark.parametrize("prob", [0.5, 1])
 @pytest.mark.parametrize("K", [32, 17])
 @pytest.mark.parametrize("M", [30, 17])
 @pytest.mark.parametrize("L", [30, 17])
-def test_sddmm_csr(L, M, K):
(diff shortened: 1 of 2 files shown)


tests/test_custom_ops.py
```diff
@@ -158,14 +158,14 @@ def test_sddmm_sputnik(device):
 
 
 @cuda_only
+@pytest.mark.parametrize("prob", [0.5, 1])
```
blefaudeux, line 161: thanks for adding this, nice catch

blefaudeux (approved):
LGTM, thanks a lot @fmassa for diving in and the very quick fix

blefaudeux:
the mypy error is fixed on main, will be fine on landing

PR #301, opened 2022-05-10
Remove double dropout in global attention

fmassa:
## What does this PR do?

`scaled_dot_product_attention` already handles dropout internally.
In
https://github.com/facebookresearch/xformers/blob/8a0507086e5fca99da752258b3c2b56b233ceaa4/xformers/components/attention/global_tokens.py#L119
we are applying dropout in the output of the attention block, which is not intended.

This PR fixes it by removing the extra dropout


blefaudeux (approved):
nice catch, thanks !

PR #308, opened 2022-05-23
Fix pickling of nystrom

### LINKED ISSUE #307: Cannot pickle nystrom based model
When I `torch.save` my model with `nystrom` attention I get the following pickle error:

```py
AttributeError: Can't pickle local object 'get_avg_pool.<locals>.avg_pool'
```

I believe coming from this function:

https://github.com/facebookresearch/xformers/blob/9232b2d27a775a43173e2fd86c03251ab64f7ede/xformers/components/attention/nystrom.py#L60

kashif:
## What does this PR do?
Fixes #307.

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/xformers/components/attention/nystrom.py
+++ b/xformers/components/attention/nystrom.py
@@ -57,24 +57,28 @@ class NystromSelfAttentionConfig(AttentionConfig):
     use_razavi_pinverse: Optional[bool]
 
 
-def get_avg_pool(n: int):
-    def avg_pool(x: torch.Tensor):
+class AvgPool:
+    def __init__(self, n: int):
+        super().__init__()
+        self.n = n
+
+    def __call__(self, x: torch.Tensor):
         # Average independently for every segment in the sequence dimension
         seq_len = x.shape[1]
         head_dim = x.shape[2]
-        segments = seq_len // n
+        segments = seq_len // self.n
 
         # Dimensions are a match
-        if seq_len % n == 0:
+        if seq_len % self.n == 0:
             return x.reshape(
                 -1,
-                n,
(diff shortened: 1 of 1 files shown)


blefaudeux:
hey @kashif, thanks for this PR, looks great ! mypy would just need a type annotation to be happy, else LGTM !

xformers/components/attention/nystrom.py
```diff
@@ -57,24 +57,28 @@ class NystromSelfAttentionConfig(AttentionConfig):
     use_razavi_pinverse: Optional[bool]
 
 
-def get_avg_pool(n: int):
-    def avg_pool(x: torch.Tensor):
+class AvgPool:
```
blefaudeux, line 60: I think that mypy is right, this is missing the nn.Module inheritance probably

`class AvgPool(nn.Module):`
should be enough to fix ?

kashif:
one more thing to keep in mind is that `seq_len` needs to be somewhat bigger than the number of landmarks else you end up with 0 segments... should i add an assert or something?

blefaudeux:
> one more thing to keep in mind is that `seq_len` needs to be somewhat bigger than the number of landmarks else you end up with 0 segments... should i add an assert or something?

Good point, I would add an assert indeed, somebody knowing the method would not make the mistake but it could help debug an error still

blefaudeux (approved):
Looks great, thanks again @kashif !

kashif:
thank you!

PR #331, opened 2022-06-09
Add attention mask support to blocksparse

### LINKED ISSUE #289: Support for attention masks in blocksparse layers
# 🚀 Feature

Support for attention masks in blocksparse layers.

## Motivation

It would be nice.

## Pitch

We can pre-collect indices based on the sparse layout with this code when we run `create_triton_kernels`:

```python
            #set up preddetermined indices for gather operation during masking
            self.to_gather = []
            self.broadcast_to_gather = [] #use this if broadcasting same mask across heads
            num_col_blocks=self.layout.shape[-1]
            for idx, (h, i, j) in enumerate(zip(*self.layout.nonzero(as_tuple=True))):
                #assume mask is not 2D, otherwise very expensive
                self.to_gather.append(h*num_col_blocks+j)
                self.broadcast_to_gather.append(j)

            self.to_gather = torch.tensor(self.to_gather).long().to(device)
            self.broadcast_to_gather = torch.tensor(self.broadcast_to_gather).long().to(device)
```

Then we reshape and collect indices from the attention mask in the `forward`:
```python

            #apply masks
            if att_mask is not None:# and len(att_mask.nonzero)>0:
                #shape is either (B,1,1,S) or (B,nh,1,S)
                assert att_mask.shape[0]==q.shape[0] and att_mask.shape[1] in [q.shape[1],1] and att_mask.shape[-1]==q.shape[-2]
                #reshape into blocks
                block_att_mask = att_mask.reshape(att_mask.shape[0],att_mask.shape[1]*att_mask.shape[-1]//self.block_size,self.block_size)

                #gather based on predefined self.layout
                if att_mask.shape[1]==1:
                    block_att_mask = block_att_mask[:,self.broadcast_to_gather,:]
                else:
                    block_att_mask = block_att_mask[:,self.to_gather,:]

                #broadcast across the att_mask row dimension
                sparse_att_mat += block_att_mask.unsqueeze(-2)
```

## Alternatives

We could additionally support a 2D dense attention masking by calling `triton.testing.sparsify_tensor` but this is super slow.

## Additional context

We can probably support key_padding_mask as well. I am not sure of the standard shape for this, but the code should be similar.

I am using this locally and the speed seems fine. Happy to do a more thorough verification though.

colehawkins:
## What does this PR do?
Fixes #289. Adds attention mask support to blocksparse layers.

## Before submitting

- [ Always] Did you have fun?
- [ X] Did you read the [contributor guideline](https://github.com/facebookresearch/xformers/blob/master/CONTRIBUTING.md)?
- [ X] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
  - [ ] N/A
  - [ ] N/A
- [ X] Did you write any new necessary tests?
  - [ ] N/A
  - [ ] N/A

## PR review

Tagging @blefaudeux due to discusion in #289

### DIFF
--- a/xformers/components/attention/blocksparse.py
+++ b/xformers/components/attention/blocksparse.py
@@ -12,6 +12,7 @@ import torch
 
 from xformers import _is_triton_available
 from xformers.components.attention import Attention, AttentionConfig, register_attention
+from xformers.components.attention.utils import bool_mask_to_additive
 
 if _is_triton_available:
     from triton.ops.blocksparse import matmul as blocksparse_matmul  # type: ignore
@@ -96,8 +97,14 @@ if _is_triton_available:
             self.requires_same_k_q_dimensions = True
 
             # The underlying triton op does not support per element attention mask
-            self.supports_attention_mask = False
+            self.supports_attention_mask = True
             self.supports_key_padding_mask = False
+        
+        def update_mask_type(self, mask: torch.Tensor):
+            logging.warning(
+                "Mask has to be additive. Fixing that but this slows things down"
+            )
+            mask = bool_mask_to_additive(mask)
 
         def create_triton_kernels(self, device):
             # blocksparse operators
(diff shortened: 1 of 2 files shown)


xformers/components/attention/blocksparse.py
```diff
@@ -169,13 +191,38 @@ def forward(
...
 
             # Self-attend: (B, nh, S, hs) x (B, nh, hs, S) -> (B, nh, S, S)
             # When the computations are block sparse, the matrix types change along the way:
             # - (sparse) attention matrix = (dense) Kt * (dense) Q
             q = q / math.sqrt(q.size(-1))
             sparse_att_mat = self.sparse_dot_sdd(q, k)
 
+            # apply masks
+            if att_mask is not None:
+                # mask shape is either (B,1,1,S) or (B,nh,1,S)
```
blefaudeux, line 205: in other parts of xformers (and more generally I think given their name) attention masks relate to the attention matrix, so they don't only mark key or query but are (B,S,S) or (S,S). would that work for you ?

blefaudeux:
hey @colehawkins, many thanks for this PR, it's really smart ! We've a difference in convention in the way the attention mask is defined (it looks like you were expecting a key or query mask ?), but out of principle it should work after a small change, not a super big deal. I hope that works for you, let me know if you would like to add key or query masking instead (it's also possible, some attentions support that already)

SkyAndCloud:
@blefaudeux Hi, how to implement a blocksparse attention supporting attention mask (i.e. shape SxS)? I want to implemnt a sparse attention with specific layout. However, current blocksparse attention only use the layout, which lacks the intra-block mask.

blefaudeux:
> @blefaudeux Hi, how to implement a blocksparse attention supporting attention mask (i.e. shape SxS)? I want to implemnt a sparse attention with specific layout. However, current blocksparse attention only use the layout, which lacks the intra-block mask.

Hey @SkyAndCloud seeing this a little late, sorry about that. It's a good point, it wouldn´t be too hard to add [here](https://github.com/facebookresearch/xformers/blob/main/xformers/components/attention/blocksparse.py#L179), one would need to make sure that the indices are redirected from the original full size matrix to this one though. I don't have a lot of cycles unfortunately but I could definitely help out and review, else I'll keep that in mind and may do a PR down the line ? (no fixed ETA..)

danthe3rd:
Stale PR - closing.
The files modified here will also be deprecated in a next version

Issue #18, opened 2021-10-21
[perf] Auto causal <> sparse


blefaudeux:
# 🚀 Feature
When applicable, automatically use sparse or blocksparse for causal attention. Right now this requires that people use them explicitly, even if the causal flag is passed, which means that a lot of people could miss the possible optimization.

## Motivation
Free perf on the table, can be a little complex to handle all cases, but would make sense to do it directly in xFormers.

## Pitch
Sort out the applicable cases first, and in that case defer to the sparse or blocksparse when ScaledDotProduct is used with the causal flag

## Alternatives
Warn users of this possible optim

PR #334, opened 2022-06-14
Switch to blocksparse for causal attention

yuanandonly:
## What does this PR do?
Addresses issue [here](https://github.com/facebookresearch/xformers/issues/18). Automatically switches to blocksparse when attention is causal, and mask is not sparse. 

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [x] N/A
  - [ ] N/A
  - [x] N/A

## PR review


blefaudeux:
hey @yuanandonly, thanks for the PR ! I cannot review myself but it seems that the unit test caught something valid, the Favor mechanism can expose a causal attention but blocksparse  (normal attention) should not be used here, it's just not the same mechanism. An easy fix is to limit your background optimization to the normal attention, aka `scaled_dot_product`

xformers/components/attention/core.py
```diff
@@ -216,13 +220,44 @@ def scaled_dot_product_attention(
...
     autocast_disabled = (
         _is_sparse_available
         and isinstance(att_mask, SparseCS)
         or (att_mask is not None and att_mask.is_sparse)
     )
 
+    # Check if causal is required but mask is not sparse; if fp16 or under amp context
+    switch_to_blocksparse = (
+        _is_triton_available
+        and (att_mask is not None and not att_mask.is_sparse)  # redundant?
```
blefaudeux, line 234: not redundant, it's good I think. It's possible not to have a att_mask here, and it's also possible that there's a att_mask and it's sparse, in both cases you would not want to switch to blocksparse

fmassa (commented):
Thanks for the PR @yuanandonly !

I've left a few comments, let me know what you think.

I believe test failures are due to the fact that you return a 4d tensor instead of a 3d one.

Also, could you add a benchmark script that checks the speed of using blocksparse vs the default case, and share the results here?

tests/test_core_attention.py
```diff
@@ -109,3 +109,52 @@ def test_amp_attention_sparsecs(device):
...
+
+    # Check that a switch to blocksparse is only triggered by causal flag
+    with torch.cuda.amp.autocast():
+        r_custom = scaled_dot_product_attention(a, a, a, m_custom)
+        r_sparse = scaled_dot_product_attention(a, a, a, m_sparse)
+        r_att_mask = scaled_dot_product_attention(a, a, a, m_att_mask)
+
+    r_att_mask = r_att_mask.squeeze()
+
+    expected_device = torch.float32
```
dianaml0, line 138: nit: expected_dtype?

tests/test_core_attention.py
```diff
@@ -109,3 +109,52 @@ def test_amp_attention_sparsecs(device):
 
     expected_device = torch.float32
     assert r.dtype == expected_device
+
+
+@pytest.mark.parametrize("device", ["cuda"])
+@pytest.mark.parametrize("data_type", [torch.float16, torch.float32])
+def test_switch_blocksparse(device, data_type):
```
dianaml0, line 116: An additional test for when sequence length is not divisible by block size could be useful

xformers/components/attention/core.py
```diff
@@ -216,13 +220,44 @@ def scaled_dot_product_attention(
...
+        and (isinstance(att_mask, AttentionMask) and att_mask.is_causal)
+        and (q.dtype == torch.float16 or torch.is_autocast_enabled())
+    )
+
+    if switch_to_blocksparse:
+        drop_prob = dropout.p if isinstance(dropout, torch.nn.Dropout) else 0.0
+        seq = q.shape[-2]
+        block_size = 16  # default
+        blocks = seq // block_size
```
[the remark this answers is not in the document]
yuanandonly: Currently it says in `blocksparse.py` that  
>         .. warning: for now, the sequence (context) length has to be a power of two. This constraint could
>             be relaxed in the future.
> 
>         .. warning: the block size has to be picked from [16, 32, 64]. Some speed is gained from bigger blocks.
>             It is of course possible to reproduce coarser patterns given these primitives, as the user sees fit.

Assuming that the sequence length is always >= 64, wouldn't this constraint mean that sequences are always divisible by block length?

@dianaml0 would a test for this still be useful? I'm happy to write one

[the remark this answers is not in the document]
dianaml0: True but then we should assert the sequence length meets those conditions before switching to the blocksparse implementation. The test could then check that for certain sequence lengths we don't switch to the blocksparse implementation.

tests/test_core_attention.py
```diff
@@ -109,3 +109,52 @@ def test_amp_attention_sparsecs(device):
...
+        r_sparse = scaled_dot_product_attention(a, a, a, m_sparse)
+        r_att_mask = scaled_dot_product_attention(a, a, a, m_att_mask)
+
+    r_att_mask = r_att_mask.squeeze()
+
+    expected_device = torch.float32
+    assert r_sparse.dtype == expected_device
+
+    if r_custom.dtype == r_att_mask.dtype:
+        assert torch.allclose(r_custom, r_att_mask, atol=1e-6, rtol=1e-3)
```
yuanandonly, line 142: @blefaudeux @fmassa @dianaml0 In this test I had to increase the tolerances of `torch.allclose()` from the default `atol=1e-8` and `rtol=1e-5` pretty significantly for the assert to pass. Something similar, `assert_almost_equal()`, from is used [here](https://github.com/facebookresearch/xformers/blob/main/tests/test_triton_blocksparse.py#L204) to test parity between standard SDP attention and blocksparse attention.

Is this difference between SDP attention and blocksparse acceptable in this situation? I.e., we're silently switching to blocksparse as of now, but should we inform the user? Or are there any other steps I should take in the code

blefaudeux: we can force pytorch to use tf32 in that case also, else I think that it's fine if the tolerance relaxation is limited to this case ?

yuanandonly: @fmassa @dianaml0 Just pushed what I currently have as my benchmark file. It's a WIP so little messy, but its there in case you guys want to take a look.

yuanandonly:
@dianaml0 @fmassa @blefaudeux 
Summary of changes since last review:
- Added benchmark file
- Added LRU cache to save blocksparse objects
- Added more tests
- Added plots to benchmark.md

Do i need to update the changelog or add any other documentation? Thanks!

BENCHMARKS.md
```diff
@@ -127,3 +127,13 @@ __Some results:__
 - Estimated flops and peak memory are on the ListOps task, using 4 GPUs. Note that LRA is not completely well defined, in that hyperparameters and model architectures can vary (should the same architecture be used everywhere ? Similar hyperparams ?). This could be improved in the future, but in the meantime one should probably not read too much into small differences for some tasks, probably not meaningful.
 
 _Note_: The estimated flops currently miss accounting for many operators, and are almost certainly an undercount. See issue [#154](https://github.com/fairinternal/xformers/issues/154)
+
+
+## Casual Attention Blocksparse Optimization
```
blefaudeux, line 132: typo ?

xformers/components/attention/core.py
```diff
@@ -210,19 +215,94 @@ def scaled_query_key_softmax(
     return att
 
 
+# 128 is default maxsize
+@lru_cache(maxsize=128)
```
blefaudeux, line 219: nice, I did not think of that

xformers/components/attention/core.py
```diff
@@ -210,19 +215,94 @@ def scaled_query_key_softmax(
...
+            v = v.unsqueeze(1)
+        else:
+            q = q.unsqueeze(0)
+            k = k.unsqueeze(0)
+            v = v.unsqueeze(0)
+
+    blocksparse_attention = _retrieve_blocksparse(layout_heads, seq_len, block_size)
+    # Dropout is a no-op in evaluation mode
+    if isinstance(dropout, torch.nn.Dropout):
+        blocksparse_attention.attn_drop = dropout
```
blefaudeux, line 262: oh it would have been nice to fuse that (fuse the dropout op with the softmax for instance, not make it another call), I'm actually a little surprised that it's not the case already.. for another day

xformers/components/attention/core.py
```diff
@@ -210,19 +215,94 @@ def scaled_query_key_softmax(
...
+
+    # Switch only if sequence length is divisible by block size
+    # Blocksparse requires the same dimensions for K and Q for now
+    seq_len = q.shape[-2]
+    if (
+        switch_to_blocksparse
+        and not seq_len % block_size
+        and q.shape[-2] == k.shape[-2]
+    ):
+        # print("switching to blocksparse...")
```
blefaudeux, line 303: nit, dead code ?

xformers/components/attention/core.py
```diff
@@ -210,19 +215,94 @@ def scaled_query_key_softmax(
...
+        and (isinstance(att_mask, AttentionMask) and att_mask.is_causal)
+        and (q.dtype == torch.float16 or torch.is_autocast_enabled())
+    )
+
+    # Switch only if sequence length is divisible by block size
+    # Blocksparse requires the same dimensions for K and Q for now
+    seq_len = q.shape[-2]
+    if (
+        switch_to_blocksparse
+        and not seq_len % block_size
```
blefaudeux, line 300: nit, but could these two conditions be part of the tests just above, to decide whether to switch to blocksparse or not ? I think that it makes the flow a little easier to follow, there's a conditional branch and you can see all the factors in one place

blefaudeux:
looks great to me, thanks a lot @yuanandonly ! Small nits here (feel free to dispute), and I would definitely update the changelog with this as I think it can be significant perf wise for all GPT like workloads. Also letting @dianaml0 and @fmassa give the green light, but thanks already for the very thorough PR

blefaudeux:
Oh, and the mypy issue in the CI just require a rebase onto current main I think

xformers/components/attention/core.py
yuanandonly: Thanks!

tests/test_core_attention.py
```diff
@@ -109,3 +110,73 @@ def test_amp_attention_sparsecs(device):
...
+    # Check that a switch to blocksparse is only triggered by causal flag
+    with torch.cuda.amp.autocast():
+        r_custom = scaled_dot_product_attention(a, a, a, m_custom)
+        r_sparse = scaled_dot_product_attention(a, a, a, m_sparse)
+        r_att_mask = scaled_dot_product_attention(a, a, a, m_att_mask)
+
+    expected_device = torch.float32
+    assert r_sparse.dtype == expected_device
+
+    if r_custom.dtype == r_att_mask.dtype:
```
dianaml0, line 140: How often is this not the case?

xformers/components/attention/core.py
```diff
@@ -210,19 +215,94 @@ def scaled_query_key_softmax(
...
 
+# 128 is default maxsize
+@lru_cache(maxsize=128)
+def _retrieve_blocksparse(
+    num_heads: int, seq_len: int, block_size: int
+) -> BlockSparseAttention:
+    # Checks if blocksparse object exists in cache
+
+    blocks = seq_len // block_size
+    print("Made uncached blocksparse")
```
dianaml0, line 226: You could use logging.info here instead of print

xformers/components/attention/core.py
```diff
@@ -210,19 +215,94 @@ def scaled_query_key_softmax(
...
+    return att
+
+
 def scaled_dot_product_attention(
     q: torch.Tensor,
     k: torch.Tensor,
     v: torch.Tensor,
     att_mask: Optional[Union[AttentionMask, "SparseCS", torch.Tensor]],
     dropout: Optional[torch.nn.Module] = None,
+    block_size=128,
```
dianaml0, line 279: You can add the type here

tests/test_core_attention.py
yuanandonly: When the inputs are fp16, the datatypes for normal sdp attention and blocksparse are both the same (fp16). But when the inputs are fp32, they are different where blocksparse attention returns fp32 and sdp attention always returns an fp16 tensor. I looked into this a while ago and found that it had to do with pytorch matmuls, not completely sure though ([this](https://stackoverflow.com/questions/70650236/why-does-pytorch-matmul-get-different-results-when-executed-on-cpu-and-gpu) might be relevant)

dianaml0: What if you cast the results to fp16 or fp32 and compare?

yuanandonly: Good point, implemented!

xformers/components/attention/core.py
fmassa: In general printing in library code can be quite annoying, specially if it happens often during training.

It's ok to leave it like this for now (as the overhead of creating the `BlockSparseAttention` object is high), but in the long run it would be good to remove this print

xformers/components/attention/core.py
```diff
@@ -210,19 +215,94 @@ def scaled_query_key_softmax(
...
+    block_size: int = 128,
+) -> torch.Tensor:
+
+    orig_dim = q.dim()
+    seq_len = q.shape[-2]
+    # Layout head dimension: 1 or batch size (q.shape[0])
+    layout_heads = 1
+
+    # TODO perhaps add functionality to pad qkv if sequence length is not divisible by block size?
+    assert seq_len % block_size == 0, "Sequence length must be divisible by block size"
```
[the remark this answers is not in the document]
fmassa: I think in the future we should just have a custom set of operators (`sddmm` / `softmax` / spmm`) which work on causal structure. This would enable for fast execution without being dependent on block sizes, and should be faster than blocksparse as we wouldn't need to index on a set of (unused) indices

xformers/components/attention/core.py
```diff
@@ -210,19 +215,94 @@ def scaled_query_key_softmax(
...
+    # Dropout is a no-op in evaluation mode
+    if isinstance(dropout, torch.nn.Dropout):
+        blocksparse_attention.attn_drop = dropout
+    else:
+        blocksparse_attention.attn_drop = torch.nn.Dropout(0.0)
+    att = blocksparse_attention(q, k, v)
+
+    # Reshape attention (B, nh, S, hs) back to (N, S, hs)
+    if orig_dim == 3:
+        return reshape_heads(att, *att.size())
```
fmassa, line 269: nit: this could be simplified with `attn.flatten(0, 1)`.

xformers/components/attention/utils.py
```diff
@@ -106,3 +106,13 @@ def bool_mask_to_additive(
...
+
+
+# (B, S, D) to (B, S, nh, hs)
+def split_heads(t: torch.Tensor, B: int, nH: int, S: int, Hs: int):
+    return t.view(B, nH, S, Hs)
+
+
+# (B, nh, S, hs) back to (N, S, hs)
+def reshape_heads(t: torch.Tensor, B: int, nH: int, S: int, Hs: int):
+    return t.view(B * nH, S, Hs)
```
fmassa, line 118: nit: looks like `split_heads` is not used anymore, and given that `reshape_heads` can be simplified as `t.flatten(0, 1)` in the main code, I think we could remove those two functions

xformers/components/attention/core.py
fmassa: I think it would be a good cleanup to integrate those changes in the same `switch_to_blocksparse` (at least in the future)

fmassa (approved):
Thanks for the PR, this is looking great!

For a follow-up PR (so that we don't take too long to get the PR merged), it would be good to address some of the comments that were left by @blefaudeux @dianaml0 and myself, as I believe it could make the code a bit simpler.

PR #348, opened 2022-07-03
[fix] Layer positions labelling & layernorm

### LINKED ISSUE #347: [factory] First and last layers are not properly tagged -> superfluous layernorms
# 🐛 Bug
Related to tests with https://github.com/facebookresearch/xformers/issues/330 when generating more complex models from configs, in some cases the layers get mislabelled (several "last layer") meaning that we get a lot of extra layernorms

blefaudeux:
## What does this PR do?
Fixes  #347 

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/xformers/factory/block_factory.py
+++ b/xformers/factory/block_factory.py
@@ -175,10 +175,7 @@ class xFormerEncoderBlock(torch.nn.Module):
         # Optional patch embedding
         self.patch_emb: Optional[nn.Module] = None
 
-        if (
-            config.patch_embedding_config is not None
-            and config.layer_position.is_first()
-        ):
+        if config.patch_embedding_config is not None:
             self.patch_emb = build_patch_embedding(
                 PatchEmbeddingConfig(**config.patch_embedding_config)
             )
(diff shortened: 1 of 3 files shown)


xformers/factory/model_factory.py
```diff
@@ -149,10 +149,12 @@ def __init__(
             for i in range(config.num_layers):
                 # Label where this layer is in the stack
                 # (for instance useful for the positional encoding, or late layer norm)
-                if i > 0:
```
blefaudeux, line 152: this would count within the repeated layers, not the overall layer stack

xformers/factory/block_factory.py
```diff
@@ -175,10 +175,7 @@ def __init__(self, config: xFormerEncoderConfig, **kwargs):
         # Optional patch embedding
         self.patch_emb: Optional[nn.Module] = None
 
-        if (
```
blefaudeux, line 178: this was not respecting the config, while trying to do the right thing: if the config was asking for a patch embedding, but the layer was not first, it would not be instantiated. In retrospect I think that it's risky, not doing what the API says it will do, plus it only worked in practice because is_first() was often wrong. I think now that it's better to respect the config no matter what, and not silently diverge

dianaml0 (approved):
Nice catch!

PR #352, opened 2022-07-07
nvFuser integration for operation fusion 

yuanandonly:
## What does this PR do?
Uses AOTAutograd and NVfuser to created fused operator patterns layers
- Bias + Activation + Dropout
- Bias + Dropout + Residual
- Bias + Dropout + Residual + Layernorm

Integrated BiasActivationDropout pattern into MLP feedforward

Wrote benchmarking against pytorch eager and triton fused patterns 

## Before submitting

- [X] Did you have fun?
- [X] Did you read the [contributor guideline](https://github.com/facebookresearch/xformers/blob/master/CONTRIBUTING.md)?
  - [X] N/A
  - [X] N/A
  - [X] N/A
  - [] N/A

## PR review

### DIFF
--- a/tests/test_global_attention.py
+++ b/tests/test_global_attention.py
@@ -11,6 +11,9 @@ from xformers.components.attention import GlobalAttention, ScaledDotProduct
 def test_global_attention():
     b, s, d = 2, 90, 40
 
+    torch.cuda.manual_seed(42)
+    torch.manual_seed(42)
+
     def test_ratio(global_attention_ratio: float):
         # Make sure that Global and Normal attention get the same results for the corresponding tokens
         a = torch.rand(b, s, d)


xformers/components/nvfuser/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+    generate_matching_config,
+    get_registry_decorator,
+    import_all_modules,
+)
+
+from .base import Fused, FusedConfig  # noqa
+
+# CREDITS: Classy Vision registry mechanism
+
+FUSED_REGISTRY: Dict[str, Any] = {}
```
dianaml0, line 21: Perhaps rename this to NVFUSED_REGISTRY? TO distinguish from triton fused for example. But not sure we need this?

xformers/components/nvfuser/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+FUSED_CLASS_NAMES: Set[str] = set()
+
+
+def build_fused(config: Union[Dict[str, Any], FusedConfig]):
```
dianaml0, line 25: I'm not sure that we need this, we should default to a nvfused operation in certain cases or have a flag for it, but I don't think the user should specify a specific config for the same operation

xformers/components/nvfuser/bias_act_dropout.py
```diff
@@ -0,0 +1,93 @@
...
+    activation: Activation, prob: float, x: torch.Tensor, bias: torch.Tensor
+) -> torch.Tensor:
+    if bias is not None:
+        x = torch.add(x, bias)
+    y = activation(x)
+    return torch.nn.functional.dropout(y, prob) if prob > 0.0 else y
+
+
+@register_fused("fused_bias_activation_dropout", FusedBiasActivationDropoutConfig)
+class FusedBiasActivationDropout(Fused):
```
dianaml0, line 36: You can look at triton fused layer norm as an example [here](https://github.com/facebookresearch/xformers/blob/main/xformers/triton/layer_norm.py#L166)

xformers/components/nvfuser/bias_act_dropout.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    A layer which fuses the computation of Dropout(Activation(x))
```
dianaml0, line 38: And bias?

xformers/components/nvfuser/base.py
```diff
@@ -0,0 +1,47 @@
...
+        *args,
+        **kwargs,
+    ):
+        super().__init__()
+
+        # This feedforward requires a CUDA accelerator
+        self.requires_cuda = False
+
+        # This feedforward requires a context length which is squared, often due to 2D pooling
+        self.requires_squared_context = False
```
dianaml0, line 37: Is this used?

xformers/components/feedforward/mlp.py
```diff
@@ -34,13 +38,27 @@ def __init__(
...
-        self.mlp = nn.Sequential(
-            nn.Linear(dim_model, hidden_layer_multiplier * dim_model, bias=bias),
-            build_activation(activation),
-            nn.Dropout(dropout),
-            nn.Linear(hidden_layer_multiplier * dim_model, dim_model, bias=bias),
-            nn.Dropout(dropout),
-        )
+        dim_mlp = hidden_layer_multiplier * dim_model
+
+        if activation in {Activation.ReLU, Activation.GeLU} and _is_functorch_available:
```
dianaml0, line 43: Would be great to validate that the two code paths yeild the same result

xformers/components/nvfuser/bias_act_dropout.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@dataclass
+class FusedBiasActivationDropoutConfig(FusedConfig):
+    p: float
+    activation: Activation
+    bias_shape: Optional[int]
+
+
+def _fn(
```
dianaml0, line 26: As a followup we can experiment with other fn's and see if we get any results

xformers/components/feedforward/mlp.py
```diff
@@ -34,13 +38,27 @@ def __init__(
...
+            self.mlp = nn.Sequential(
+                nn.Linear(dim_model, hidden_layer_multiplier * dim_model, bias=False),
+                FusedBiasActivationDropout(
```
dianaml0, line 46: nit but may be a bit cleaner if we just define the FusedBiasActiationDropout in the if/else and then build the mlp after?

xformers/benchmarks/benchmark_nvfuser.py
```diff
@@ -0,0 +1,134 @@
...
+    total = 2 * a.numel() * a.element_size()
+
+    if bias:
+        # Read the bias, ideally only once
+        total += a.shape[-1] * a.element_size()
+
+    return total * 1e-9 / (ms * 1e-3)
+
+
+def bench_dropout(bias: bool, backward: bool, activation: Activation):
```
dianaml0, line 40: nit: rename?

dianaml0 (commented):
Thanks for the initial work here @yuanandonly! Left a few comments

blefaudeux:
Nice, and the perf is super good ! This is pretty great @yuanandonly, thanks already ! 

I would agree with @dianaml0 and it will be simpler for you actually, probably better not to register a new layer type ? It can be automatically (conditionally on nvfuser being present and in the cases which are just faster) applied to the existing layers, that would be even better I feel -less complexity and automatically improving the situation for all users-. You can still expose a macro flag, in the form of a bool that users can flip -possibly at runtime-, for debugging for instance

fmassa (commented):
Thanks for the PR!

On top of what Diana and Benjamin mentioned, I have also a couple of comments.

Chillee (commented):
Usage of AOTAutograd looks fine to me.

I do wonder about the API though. One of the nice things about using AOTAutograd is that it gives the user a lot of flexibility in how to construct their function. It seems like this API sacrifices that advantage.

Also, it might be worth trying out the TorchInductor backend as well.

yuanandonly:
@Chillee What sort of flexibility are you suggesting for the API? Do you mean keeping the nvfuser component more general instead of having different components for the different fused patterns?

xformers/components/nvfuser/bias_dropout_res.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import Any, Optional
+
+import torch
+import torch.nn as nn
+from functorch.compile import memory_efficient_fusion
+
+from xformers.components import LayerNormStyle
+
+
+def _fn(
```
dianaml0, line 17: Perhaps we can pull out a general method that takes in a function and outputs the nvfused version. Then we can use that to provide these various fusions

dianaml0: It could remove some redundancy in these files

xformers/components/feedforward/mlp.py
```diff
@@ -29,18 +37,43 @@ def __init__(
         activation: Activation,
         hidden_layer_multiplier: int,
         bias: bool = True,
+        nvfuser: bool = True,
```
dianaml0, line 40: We shouldn't expose this here, just default to it if available.

xformers/components/nvfuser/bias_dropout_res_layernorm.py
```diff
@@ -0,0 +1,91 @@
...
+    def forward(self, x: torch.Tensor, res: torch.Tensor) -> torch.Tensor:
+        # Train/inference
+        p = self.p if self.training else 0.0
+
+        # Catch possible type or device mismatch
+        self.norm = self.norm.to(dtype=x.dtype, device=x.device)
+        if self.bias is not None and self.bias.dtype != x.dtype:
+            self.bias = nn.Parameter(
+                torch.zeros_like(self.bias, dtype=x.dtype, device=x.device)
+            )
```
fmassa, line 81: This shouldn't be in the `forward`. Parameter initialization should happen in the constructor, and the `norm` cast should be handled naturally given that `norm` is a `nn.Module`

xformers/components/nvfuser/bias_dropout_res.py
```diff
@@ -0,0 +1,68 @@
...
+        p: float,
+        bias_shape: Optional[int] = None,
+    ) -> None:
+        super().__init__()
+
+        self.p = float(p)
+        self.requires_residual = True
+
+        self.bias = (
+            nn.Parameter(torch.zeros(bias_shape, device=torch.device("cuda")))
```
fmassa, line 42: same comment in here about `cuda`

tests/test_nvfuser.py
```diff
@@ -0,0 +1,172 @@
...
+    from xformers.components.nvfuser import (
+        NVFusedBiasActivationDropout,
+        NVFusedBiasDropoutRes,
+        NVFusedBiasDropoutResLayerNorm,
+    )
+    from xformers.components.nvfuser.utils import build_nvfused
+except ImportError as e:
+    logging.warning(f"Functorch is not available in test_nvfuser.py. \nError {e}")
+
+FUSED_PATTERNS = (
```
dianaml0, line 32: Looks like the test is erroring here

tests/test_nvfuser.py
```diff
@@ -0,0 +1,172 @@
...
+
+import xformers
+from xformers.components import Activation, LayerNormStyle
+from xformers.components.feedforward import build_feedforward
+
+_gpu_available = torch.cuda.is_available()
+
+xformers._is_functorch_available = True
+
+try:
```
dianaml0, line 22: `pip install functorch` should probably be added [here](https://github.com/facebookresearch/xformers/blob/main/.circleci/config.yml)?

tests/test_nvfuser.py
```diff
@@ -0,0 +1,172 @@
...
+    p: float,
+    layer_norm_style: LayerNormStyle,
+):
+
+    if (
+        fused_pattern != NVFusedBiasDropoutResLayerNorm
+        and layer_norm_style == LayerNormStyle.Post
+    ):
+        pytest.skip(
+            "Layer norm style doesn't apply, the same relevant params already tested once."
```
dianaml0, line 95: How come we skip here? Should this error out if a user tries it?

tests/test_nvfuser.py
```diff
@@ -0,0 +1,172 @@
...
+        fused.train().cuda()
+        nvfused_res = fused(x, x) if fused.requires_residual else fused(x)
+        fused.cpu()
+        torch_res = (
+            fused(x_cpu, x_cpu).cuda()
+            if fused.requires_residual
+            else fused(x_cpu).cuda()
+        )
+
+        # Check if operation was actually fused
```
dianaml0, line 116: Nice!

tests/test_nvfuser.py
```diff
@@ -0,0 +1,172 @@
...
+        assert isinstance(
+            nvfused_res.grad_fn, torch.autograd.function.BackwardCFunction
+        )
+
+        if p == 0.0:
+            # Check fused and unfused paths are the same
+            assert torch.allclose(torch_res, nvfused_res, atol=1e-6, rtol=1e-2)
```
dianaml0, line 123: does `torch.equal` fail?

tests/test_nvfuser.py
```diff
@@ -0,0 +1,172 @@
...
+    torch.random.manual_seed(0)
+    torch.cuda.manual_seed_all(0)
+
+    mlp = build_feedforward(test_config)
+    # Creates non-fused default MLP
+    xformers._is_functorch_available = False
+    mlp_default = build_feedforward(test_config)
+    xformers._is_functorch_available = True
+
+    if mlp.requires_cuda and not device.type == "cuda":
```
dianaml0, line 151: Shouldn't the entire test require cuda? So we can add that check there and remove this line?

tests/test_nvfuser.py
```diff
@@ -0,0 +1,172 @@
...
+
+    # Check fused pattern w/ unfused default (switch happens within NVFusedBiasActivationDropout)
+    mlp.cuda()
+    fused_res = mlp(inputs)
+
+    mlp.cpu()
+    unfused_res = mlp(inputs.cpu())
+
+    if p == 0.0:
+        assert torch.allclose(unfused_res.cuda(), fused_res, atol=1e-6, rtol=1e-2)
```
dianaml0, line 166: Same here, does `torch.equal` not work?

xformers/components/feedforward/mlp.py
```diff
@@ -9,15 +9,23 @@
...
+    )
+
 from . import register_feedforward
 
 
 @dataclass
 class MlpConfig(FeedforwardConfig):
     hidden_layer_multiplier: int
+    bias: bool
+    nvfuser: bool
```
dianaml0, line 28: Do we need this? Don't we just default to the nvfused implementation when functorch is available?

tests/test_nvfuser.py
```diff
@@ -0,0 +1,172 @@
...
+
+    # Check fused pattern w/ unfused default (switch happens within MLP)
```
dianaml0, line 168: Should there be an assert here?

xformers/__init__.py
```diff
@@ -10,8 +10,11 @@
 # Please update the doc version in docs/source/conf.py as well.
 __version__ = "0.0.12.dev"
 
-_is_sparse_available = True
-_is_triton_available = torch.cuda.is_available()
+_is_sparse_available: bool = True
```
dianaml0, line 13: Nice catch

xformers/components/feedforward/mlp.py
```diff
@@ -33,14 +41,36 @@ def __init__(
...
-            nn.Linear(dim_model, hidden_layer_multiplier * dim_model, bias=bias),
-            build_activation(activation),
-            nn.Dropout(dropout),
-            nn.Linear(hidden_layer_multiplier * dim_model, dim_model, bias=bias),
-            nn.Dropout(dropout),
-        )
+        dim_mlp = hidden_layer_multiplier * dim_model
+        # check if fused Bias Activation Dropout is applicable
+        if xformers._is_functorch_available:
+            self.requires_cuda = True
```
dianaml0, line 47: should we instead check if device is cuda and functorch is available?

xformers/components/nvfuser/bias_act_dropout.py
```diff
@@ -0,0 +1,72 @@
...
+    def init_weights(self, *args, **kwargs):
+        with torch.no_grad():
+            if self.bias is not None:
+                self.bias.fill_(0.0)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        # Train/inference
+        p = self.p if self.training else 0.0
+
+        # TODO perf check, is slower than pytorch for small buffers, bypassing it in that case
```
dianaml0, line 64: remove?

tests/test_nvfuser.py
yuanandonly: This is because the layernormstyle argument doesn't apply to the other patterns, so when layer_norm_style == LayerNormStyle.Pre all the possible configurations have already been tested once

xformers/components/nvfuser/bias_dropout_res.py
```diff
@@ -0,0 +1,66 @@
...
+                self.bias.fill_(0.0)
+
+    def forward(self, x: torch.Tensor, residual: torch.Tensor) -> torch.Tensor:
+        # Train/inference
+        p = self.p if self.training else 0.0
+
+        # TODO perf check, is slower than pytorch for small buffers, bypassing it in that case
+
+        # Catch a non-cuda setup, fallback to pytorch
+        if not x.is_cuda:
```
dianaml0, line 61: Since we have this, then we don't necessarily need the cuda check in the MLP, maybe good for clarity though

tests/test_nvfuser.py
yuanandonly: Yes it does, for some reason

xformers/components/nvfuser/bias_dropout_res.py
```diff
@@ -0,0 +1,66 @@
...
+    def init_weights(self, *args, **kwargs):
+        with torch.no_grad():
+            if self.bias is not None:
+                self.bias.fill_(0.0)
+
+    def forward(self, x: torch.Tensor, residual: torch.Tensor) -> torch.Tensor:
+        # Train/inference
+        p = self.p if self.training else 0.0
+
+        # TODO perf check, is slower than pytorch for small buffers, bypassing it in that case
```
dianaml0, line 58: remove?

tests/test_nvfuser.py
yuanandonly: It does not work

xformers/components/feedforward/mlp.py
yuanandonly: Good catch! I forgot to remove this.

xformers/components/nvfuser/bias_dropout_res_layernorm.py
```diff
@@ -0,0 +1,82 @@
...
+    a = torch.add(x, bias) if bias is not None else x
+    b = torch.nn.functional.dropout(a, prob) if prob > 0.0 else a
+    if layer_norm_style == LayerNormStyle.Pre:
+        c = norm(b)
+        return torch.add(c, residual)
+    elif layer_norm_style == LayerNormStyle.Post:
+        c = torch.add(b, residual)
+        return norm(c)
+    else:
+        raise ValueError
```
dianaml0, line 33: Add this case to testing?

xformers/components/nvfuser/bias_dropout_res_layernorm.py
```diff
@@ -0,0 +1,82 @@
...
+    def init_weights(self, *args, **kwargs):
+        with torch.no_grad():
+            if self.bias is not None:
+                self.bias.fill_(0.0)
+
+    def forward(self, x: torch.Tensor, residual: torch.Tensor) -> torch.Tensor:
+        # Train/inference
+        p = self.p if self.training else 0.0
+
+        # TODO perf check, is slower than pytorch for small buffers, bypassing it in that case
```
dianaml0, line 74: delete?

tests/test_nvfuser.py
yuanandonly: It's not possible to numerically compare the outputs here, because the different MLP objects have different random weights instantiated in the linear layers. I was unsure of whether to add any other checks

xformers/components/attention/blocksparse.py
```diff
@@ -14,6 +14,7 @@
 from xformers.components.attention import Attention, AttentionConfig, register_attention
 
 if _is_triton_available:
+
```
fmassa, line 17:
```suggestion
```

tests/test_nvfuser.py
fmassa: You can make the weights be the same so that you can compare both. You can do something like `mlp.load_state_dict(mlp_default.state_dict())` to make the weights the same.
This will ensure that we are testing everything

xformers/__init__.py
```diff
@@ -10,8 +10,11 @@
...
+_is_triton_available: bool = torch.cuda.is_available()
+
+# Set to true to utilize functorch
+_is_functorch_available: bool = False
```
fmassa, line 17: What is the desired way to let users have the nvfuser path enabled? Will they need to clone and modify xformers for that?

xformers/components/feedforward/mlp.py
```diff
@@ -33,14 +40,36 @@ def __init__(
...
+            self.mlp = nn.Sequential(
+                nn.Linear(
+                    in_features=dim_model, out_features=dim_mlp, bias=False
+                ),  # bias is handled in the next layer
+                NVFusedBiasActivationDropout(
+                    p=dropout,
+                    bias_shape=dim_mlp if bias else None,
+                    activation=activation,
+                ),
```
fmassa, line 55: The fact that the bias live in a different module means that you won't be able to easily swap to use nvfuser on a pre-trained model which used the standard layers. This is IMO a drawback which would be good to be solved.

Using this optimization means that this is not just a transparent change to the user, and the snippet like I proposed before with `state_dict` won't work out of the box.

xformers/components/nvfuser/bias_dropout_res.py
```diff
@@ -0,0 +1,65 @@
...
+            if self.bias is not None:
+                self.bias.fill_(0.0)
+
+    def forward(self, x: torch.Tensor, residual: torch.Tensor) -> torch.Tensor:
+        # Train/inference
+        p = self.p if self.training else 0.0
+
+        # Catch a non-cuda setup, fallback to pytorch
+        if not x.is_cuda:
+            print("defaulting")
```
fmassa, line 60: This is probably a leftover for debugging?

fmassa (commented):
Thanks for the changes Chris!

I've left some more comments. I think that ideally we would want the nvfuser model to be a drop-in replacement for the vanilla model, which would imply that the weight layout should be the same. I understand this might complicate a few things for now, so let's not make this a merge-blocking requirement, but it would be good to get this fixed.

xformers/__init__.py
yuanandonly: I think currently we have it as something that users will need to modify. I haven't looked into it a great deal, perhaps users who haven't cloned xformers can do something like `python -c "import xformers; xformers._is_functorch_available=True;`

tests/test_nvfuser.py
dianaml0: Does it also fail for smaller atol and rtol?

tests/test_nvfuser.py
dianaml0: So should this just be `if fused_pattern != NVFusedBiasDropoutResLayerNorm` ?

CHANGELOG.md
```diff
@@ -7,18 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## TBD
 ### Fixed
-- Removed dupliacated biases in the FusedMLP layers [#317]
+- Removed duplicated biases in the FusedMLP layers [#317]
```
dianaml0, line 10: Thanks!

HOWTO.md
```diff
@@ -12,6 +12,7 @@ Let's present here a couple of code snippets on how to solve a couple of questio
     - [Replace all attentions from an existing ViT model with a sparse equivalent ?](#replace-all-attentions-from-an-existing-vit-model-with-a-sparse-equivalent-)
     - [Some more examples](#some-more-examples)
   - [BlockSparseAttention](#blocksparseattention)
+  - [AOTAutograd and NVFuser](#aotautograd-and-nvfuser)
```
dianaml0, line 15: Maybe I would make this title more explicit? How to enable fused operations available via aotautograd/nvfuser or something like that

HOWTO.md
```diff
@@ -295,6 +296,55 @@ On a V100, with PyTorch 1.9, Triton 1.1 and xFormers 0.0.2 this reports somethin
...
+
+XFormers uses `memory_efficient_fusion` to combine sequences of fusable operations together into single fused function layers. These parts can be found [here](xformers/components/nvfuser). A notable example is [`NVFusedBiasActivationDropout`](xformers/components/nvfuser/bias_act_dropout.py), which is readily used inside the [`MLP`](xformers/components/feedforward/mlp.py) feedforward component.
+
+A benchmark of these fused patterns across some representative shapes shows significant speed increases compared to the unfused, Pytorch eager approach&mdash;up to 3.5x speedup for the forward pass and 2.2x for the forward and backward passes together. On average, peak memory usage of fused patterns is also lower, although we see some infrequent cases of up to 1.6x Pytorch peak memory usage on larger shapes. We also see better overall performance against our implementation of fused Bias, Activation, and Dropout using Triton ([see](xformers/triton/dropout.py)) as well. Full benchmark plots can be found [here](docs/plots/nvfuser/).
+
+Below is a simple example use case of AOT Autograd.
+
+```python
```
dianaml0, line 309: Here may be useful to focus more on how to use these within xFormers rather than how to implement them. For example, how to enable defaulting to nfused ops within the MLP.

README.md
```diff
@@ -78,12 +78,24 @@ Some advices related to building these CUDA-specific components, tentatively adr
 
 <details><summary> Triton </summary><p>
 
-Some parts of xFormers use [Triton](http://www.triton-lang.org), and will only expose themselves if Triton is installed, and a compatible GPU is present (nVidia GPU with tensor cores). If Triton was not installed as part of the testing procedure, you can install it directly by running `pip install triton`. You can optionally test that the installation is successful by running one of the Triton-related benchmarks, for instance `python3 xformers/benchmarks/benchmark_triton_softmax.py`
+Some parts of xFormers use [Triton](http://www.triton-lang.org), and will only expose themselves if Functorch is installed, and a compatible GPU is present (nVidia GPU with tensor cores). If Triton was not installed as part of the testing procedure, you can install it directly by running `pip install triton`. You can optionally test that the installation is successful by running one of the Triton-related benchmarks, for instance `python3 xformers/benchmarks/benchmark_triton_softmax.py`
```
dianaml0, line 81: Revert this change?

README.md
```diff
@@ -78,12 +78,24 @@ Some advices related to building these CUDA-specific components, tentatively adr
...
 
 Triton will cache the compiled kernels to `/tmp/triton` by default. If this becomes an issue, this path can be specified through the `TRITON_CACHE_DIR` environment variable.
 
 </p></details>
 
+<details><summary> NVFuser </summary><p>
+
+Some parts of xFormers use AOT Autograd from the [functorch](https://pytorch.org/functorch/stable/) library, and will only expose themselves if functorch is installed, and a compatible GPU is present (nVidia GPU with tensor cores). If functorch was not installed as part of the testing procedure, you can install it directly through pip.
```
dianaml0, line 89: I don't think tensor cores are required for using functorch

README.md
```diff
@@ -78,12 +78,24 @@ Some advices related to building these CUDA-specific components, tentatively adr
...
+
+  ```bash
+  pip install functorch
+  ```
+
+ Once installed, set the flag `_is_functorch_available : bool = True` in `xformers/__init__.py`. You can optionally test that the installation is successful by running one of the functorch-related benchmarks `python3 xformers/benchmarks/benchmark_nvfuser.py`
```
dianaml0, line 95: I would add how to modify the flag if importing the library

PR #354, opened 2022-07-21
Allow BFloat16 for blocksparse

### LINKED ISSUE #353: Blocksparse forced to use FP16 when BF16 is supported (for triton v2)
# 🐛 Bug

Within the blocksparse implementation,`q,k,v` are forced to half: https://github.com/facebookresearch/xformers/blob/main/xformers/components/attention/blocksparse.py#L171

After #231 I think blocksparse does support bfloat16 (I tried locally and seems it works). If there are no objections, we should modify the constraint to allow bfloat16. Unsure what the correct approach would be (could use the AMP region to infer the type, or the input type).

SeanNaren:
## What does this PR do?
Fixes #353 

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Rotary embeddings respecting input types [#326]
 - Poolformer style instantiating useless projection layers [#349]
 - Fix layer position not being properly tracked, causing extra layernorms for programmatic xformers [#348]
+- Blocksparse respecting native type [#354]
 
 ### Added
 - Four blocksparsity layouts from DeepSpeed [#320]
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -27,7 +27,7 @@ hydra-core >= 1.1
 fairscale >= 0.4.5
 
 # Dependency for fused layers, optional
-triton == 2.0.0.dev20220701
+triton == 2.0.0.dev20220720
 
 # Dependencies for fused layers using FuncTorch, optional
 git+https://github.com/pytorch/functorch@v0.2.0
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -27,7 +27,7 @@ hydra-core >= 1.1
 fairscale >= 0.4.5
 
 # Dependency for fused layers, optional
-triton == 2.0.0.dev20220701
+triton == 2.0.0.dev20220721
 
 # Dependencies for fused layers using FuncTorch, optional
 git+https://github.com/pytorch/functorch@v0.2.0
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -27,7 +27,7 @@ hydra-core >= 1.1
 fairscale >= 0.4.5
 
 # Dependency for fused layers, optional
-triton == 2.0.0.dev20220721
+triton == 2.0.0.dev20220701
 
 # Dependencies for fused layers using FuncTorch, optional
 git+https://github.com/pytorch/functorch@v0.2.0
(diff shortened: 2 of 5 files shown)


fmassa (commented):
Thanks for the PR!

I'm ok with the direction and would be fine merging the PR as is, but what do you think if we instead don't perform any casting whatsoever in the code? I understand that Triton won't work with fp32 (or at least it won't be efficient), but should we implicitly perform this cast here, instead of raising an error?

cc @blefaudeux for thoughts

blefaudeux:
> Thanks for the PR!
> 
> I'm ok with the direction and would be fine merging the PR as is, but what do you think if we instead don't perform any casting whatsoever in the code? I understand that Triton won't work with fp32 (or at least it won't be efficient), but should we implicitly perform this cast here, instead of raising an error?
> 
> cc @blefaudeux for thoughts

thanks for the PR @SeanNaren to begin with ! The cast (ours) is not super nice to begin with, it's mentioned in the doc in that originally blocksparse would crash on fp32, so we officially just supported fp16 and added that. I think that the fp32 case is fixed by now, and this nuked bf16 altogether which is just nicer than fp16 most of the time, so we should probably revisit that ! I can submit a PR on your branch @SeanNaren to try to remove the cast altogether, and adjust the unit tests to make sure that we cover parity testing for all types, what do you think ?

blefaudeux:
I've pushed a small addition to https://github.com/blefaudeux/xformers/tree/fix/blocksparse_bfloat16 (can be PRd onto your branch @SeanNaren ), the issue right now is that some parity tests are not passing anymore and I didn´t have time to investigate. I think that it could be related to different defaults nowadays with respect to the accumulators, triton defaults to TF32 and I think that pytorch reverted that to fp32

SeanNaren:
Thanks @fmassa @blefaudeux! getting rid of the check would be the optimal solution, @blefaudeux you should be able to push directly to my branch, feel free to do so!

blefaudeux:
I just pushed @SeanNaren, underwater these days with a new job, really sorry for the delay ! I just had to adapt a few tests (some pytorch codepaths are not bfl16 compatible), but nothing super big. I took the liberty to rebase your branch also, to remove a merge conflict, you would have to `git reset --hard origin/yourbranch` if pulling on your end

SeanNaren:
> I just pushed @SeanNaren, underwater these days with a new job, really sorry for the delay ! I just had to adapt a few tests (some pytorch codepaths are not bfl16 compatible), but nothing super big. I took the liberty to rebase your branch also, to remove a merge conflict, you would have to `git reset --hard origin/yourbranch` if pulling on your end

Appreciate it man! Without a dev machine (for now) so can't make progress here, so appreciate the assist!

blefaudeux:
Seems good except for two tests where the results are markedly different in between triton and torch, not too nice.. I'll try to update triton pip package, could be that this was fixed in betwee

blefaudeux:
I tried a couple of obvious takes, to no avail, the TL: DR is that parity with pytorch does not seem to be respected with the triton operator with bf16, with or without the tf32 accumulation and with or without the `triton.testing` tools. I'm not sure how much of this is xformers business, since this part is just a helper to use Triton's blocksparse attention, but since it's exposed here it makes sense that parity is tested. @ptillet sorry for the ping, but is that expected ? Are there some specific pip packages which we should stick to ?

blefaudeux:
fixed by https://github.com/facebookresearch/xformers/pull/528 ! Thanks @SeanNaren for putting this one up, and sorry that it was so slow to close the case, it required upgrading all the triton code here and I was a bit too busy for that..

PR #368, opened 2022-08-25
Move memory-efficient attention kernels to its own folder

fmassa:
## What does this PR do?

With additional kernels being added for the library, it might be better to move them to a single location so that it's clearer what they do.

This shouldn't change any of the functionality of the code.


fmassa:
We have been having compilation issues in functorch for some time now, so those compilation issues are unrelated

fmassa:
Merging as this is harmless change

PR #369, opened 2022-08-25
Add .contiguous() to several tensor in the sparse folder

simonrouard:
## What does this PR do?
When training a transformer that uses SparseCSRTensor, I had a bug "Expected D1.is_contiguous() to be True but got False. It happened during the evaluation time. I also had bug when the batch size was 1 or when using only 1 GPU. 
These little fixes solved my problem and they would be useful since I need to release a model that uses xformers. 
## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/xformers/sparse/_csr_ops.py
+++ b/xformers/sparse/_csr_ops.py
@@ -126,7 +126,7 @@ class _spmm(torch.autograd.Function):
         ctx, b, row_indices, values, row_offsets, column_indices, m, _transp_info
     ):
         out = torch.ops.xformers.spmm_sputnik(
-            b, row_indices, values, row_offsets, column_indices, m
+            b.contiguous(), row_indices, values, row_offsets, column_indices, m
         )
 
         ctx.save_for_backward(
@@ -149,7 +149,7 @@ class _spmm(torch.autograd.Function):
         # gradients w.r.t. values
         grad = grad.contiguous()
 
-        grad_sparse = _sddmm_func(grad, b, row_indices, row_offsets, column_indices)
+        grad_sparse = _sddmm_func(grad, b.contiguous(), row_indices, row_offsets, column_indices)
 
         (
             row_indices_t,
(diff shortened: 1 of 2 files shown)


xformers/sparse/_csr_ops.py
```diff
@@ -126,7 +126,7 @@ def forward(
         ctx, b, row_indices, values, row_offsets, column_indices, m, _transp_info
     ):
         out = torch.ops.xformers.spmm_sputnik(
-            b, row_indices, values, row_offsets, column_indices, m
+            b.contiguous(), row_indices, values, row_offsets, column_indices, m
```
fmassa, line 129: nit: if you do

```python
b = b.contiguous()
```
then you don't need to perform the `.contiguous()` call in the backward, as it will be stored directly as a contiguous tensor. This will make things slightly faster

fmassa (approved):
Thanks!

I originally kept this without the `.contiguous()` call because copying the tensors incur some overhead, but indeed it might be better here to just work instead of erroing out to users.

xformers/sparse/_csr_ops.py
```diff
@@ -148,8 +149,9 @@ def backward(ctx, grad):
 
         # gradients w.r.t. values
         grad = grad.contiguous()
+        b = b.contiguous()
```
fmassa, line 152: this is fine, but note that this is not necessary now because we cast it as contiguous in forward

xformers/sparse/csr_tensor.py
```diff
@@ -183,6 +183,8 @@ def _masked_matmul(cls, a, b, mask):
         row_indices = mask.__row_indices
         row_offsets = mask.__row_offsets
         column_indices = mask.__column_indices
+        a = a.contiguous()
+        b = b.contiguous()
```
fmassa, line 187: This is not equivalent to what you had before actually, you want to do `b.transpose().contiguous()`

fmassa (approved):
Thanks a ton!

PR #370, opened 2022-08-26
Raise early errors if submodules not available

fmassa:
## What does this PR do?

If the user forgot to perform `git submodule update --init --recursive` in their clone, they would get cryptic errors during compilation.

Now raise errors early with an explicit message.


fmassa:
Merging as this is a harmless PR

PR #384, opened 2022-09-07
AttentionFwd: Use exp2f instead of expf to gain ~5%

danthe3rd:
Taken from FlashAttention:
https://github.com/HazyResearch/flash-attention/blob/04fb198523a687f8f66e304e8555b3d62cb52a2a/csrc/flash_attn/src/fmha/softmax.h#L222-L224


fmassa (approved):
Thanks!

Do you have an idea on how much this change affects numerics?

PR #391, opened 2022-09-15
[MMA_FROM_SMEM_IT_RES] Fix padding of LSE in forward

danthe3rd:
LSE was padded with uninitialised values rather than `inf` as expected

### Stacked PR Chain: MMA_FROM_SMEM_IT_RES
| PR | Title |  Merges Into  |
|:--:|:------|:-------------:|
|#391|[MMA_FROM_SMEM_IT_RES] Fix padding of LSE in forward|**N/A**|
|#392|[MMA_FROM_SMEM_IT_RES] Copy over files from cutlass|#391|
|#393|[MMA_FROM_SMEM_IT_RES] Change how residuals are handled|#392|


fmassa (approved):
Thanks!

Would there be value in adding a test-case that would cover this?

PR #396, opened 2022-09-16
Memory Efficient Benchmarks improvements

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #443
* #442
* #440
* #439
* #438
* #434
* #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* #399
* #398
* #397
* __->__ #396
* #395

* Add a few shapes that we want to keep track of
* Include the algorithm used in the column title
* Handle OOM cases for vanilla/optimized
* Make benchmarks faster by setting `min_run_time = 0.5`
* Benchmark with q/k/v 3 different vectors - rather than k=v=q

**TEST PLAN**
```
$ python xformers/benchmarks/benchmark_mem_eff_attention.py --compare one --label two

...

[--------------- attention backward (attn_bias=<class 'xformers.ops.LowerTriangularMask'>) ----------------]
                              |  two[flshatt]  |  two[fwd_gen]  |  vanilla  |  one[flshatt]  |  one[fwd_gen]
1 threads: -------------------------------------------------------------------------------------------------
      f16 B=384, M=197, K=64  |     220.5      |                |   1706.3  |     220.9      |
      b16 B=384, M=197, K=64  |     223.7      |                |    576.7  |     223.8      |
      f32 B=384, M=197, K=64  |                |      813.2     |   1566.3  |                |      814.2
      f32 B=384, M=197, K=80  |                |     1171.7     |   1656.4  |                |
      f32 B=384, M=197, K=88  |                |     1224.9     |   1697.7  |                |     1226.1

Times are in microseconds (us).
```

PR #395, opened 2022-09-15
Fix bugs in backward. More/faster tests

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #443
* #442
* #440
* #439
* #438
* #434
* #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* #399
* #398
* #397
* #396
* __->__ #395

**TEST PLAN**
```
CUDA_VISIBLE_DEVICES=1 CUDA_LAUNCH_BLOCKING=1 time python -m pytest tests/test_mem_eff_attention.py
```

The tests run in ~22s on my machine now

PR #397, opened 2022-09-16
AttentionBW: Reverse iteration order

danthe3rd:
* #443
* #442
* #440
* #439
* #438
* #434
* #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* #399
* #398
* __->__ #397
* #396
* #395

**SUMMARY**
Iterate on queries in the inner loop


xformers/components/attention/csrc/cuda/mem_eff_attention/kernel_backward.h
```diff
@@ -400,100 +402,6 @@ struct AttentionBackwardKernel {
     };
   };
 
-  // OLD VERSION - a3f257389709
-  template <int kElementsPerAccess>
-  static __device__ void _computeDi(
```
fmassa, line 405: nit: why not taking this function and use it in a single kernel launch instead of writing the (slower) pytorch unfused ops?

fmassa (approved):
Thanks Daniel!

Approving to unblock, but it would have been good to have some more description about the perf characteristics of this change. From my understanding this PR slows things down, but enables further optimizations in the future -- a comment like that would be helpful in the future.

PR #405, opened 2022-09-20
mma_pipeline: Preload operands for GradQ/GradV on A100

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #443
* #442
* #440
* #439
* #438
* #434
* #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* __->__ #405
* #404
* #403
* #399
* #398
* #397
* #396
* #395

**PERFORMANCE**
<details>
<summary>P100/V100 fw</summary>

```
[--------------------------------------------------------------------------------------------------- attention (attn_bias=<class 'NoneType'>) ---------------------------------------------------------------------------------------------------]
                                                     |  9_mma_pipeline_e6a069a9  |  3_benchmarks_6ca785d9  |  vanilla   |  4_reverseIter_d5ca60bd  |  5_64x128_f3a3fdc2  |  6_outputrf_6e2b2934  |  7_prefetch_ed1cc988  |  8_ep_pipeline_5e166ece
1 threads: ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  (Quadro_GP100)          f16 B=16, M=4096, K=40     |           10723.8         |          10692.3        |   13552.2  |          10699.9         |        10601.6      |         10663.0       |         10719.6       |          10645.8
                          f32 B=16, M=4096, K=40     |            9549.6         |           9467.5        |   17016.0  |           9589.1         |         9542.4      |          9560.2       |          9550.8       |           9619.8
                          f16 B=16, M=16384, K=40    |          159276.6         |         159928.8        |            |         161344.5         |       160913.5      |        161142.8       |        159579.8       |         159865.1
                          f32 B=16, M=16384, K=40    |          144554.5         |         144780.6        |            |         145508.0         |       144733.1      |        144795.3       |        144050.0       |         144648.9
                          f16 B=160, M=4096, K=128   |          235078.7         |         236423.3        |  190506.4  |         237019.9         |       234904.2      |        235344.9       |        235237.9       |         235100.1
                          f32 B=160, M=4096, K=128   |          225691.2         |         225535.9        |            |         224368.2         |       225106.3      |        226960.8       |        224072.3       |         225576.0
                          f16 B=160, M=8192, K=128   |          963693.3         |         958186.1        |            |         963277.1         |       963158.5      |        964223.6       |        959088.5       |         965950.2
[...]

PR #403, opened 2022-09-19
Prefetch output tensor in epilogue

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #443
* #442
* #440
* #439
* #438
* #434
* #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* __->__ #403
* #399
* #398
* #397
* #396
* #395

PR #399, opened 2022-09-16
BW: Store dK/dV in registers when possible

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #443
* #442
* #440
* #439
* #438
* #434
* #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* __->__ #399
* #398
* #397
* #396
* #395

PR #398, opened 2022-09-16
Set blockSize dynamically to 64x64 or 64x128

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #443
* #442
* #440
* #439
* #438
* #434
* #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* #399
* __->__ #398
* #397
* #396
* #395

PR #411, opened 2022-09-21
outputf16: Change forward's output to match input dtype

danthe3rd:
* #443
* #442
* #440
* #439
* #438
* #434
* #429
* #426
* #421
* #415
* #412
* __->__ #411
* #433
* #432
* #405
* #404
* #403
* #399
* #398
* #397
* #396
* #395

- We still accumulate in f32 in gmem (or RF) when `scalar_t` is f16 (for numerical stability reasons)
- Added tests cases to trigger gmem accumulation (K>128 && M>128)
- Merged the 2 epilogue we have (so now we are also using the pipelined epiloque in forward)
- Bonus: this makes backward faster because we avoid 2 casts (output+grad)

### DIFF
--- a/xformers/components/attention/csrc/cuda/mem_eff_attention/attention_forward_generic.cu
+++ b/xformers/components/attention/csrc/cuda/mem_eff_attention/attention_forward_generic.cu
@@ -143,11 +143,12 @@ efficient_attention_forward_generic(
                       value.stride(1) % Kernel::kAlignmentV == 0,
                       "value is not correctly aligned");
 
-                  res = at::zeros(
+                  res = at::empty(
                       {B, M, K},
                       query.options().dtype(
                           TypeTraits<
                               typename Kernel::output_t>::atScalarType()));
+
                   // NOTE: Should be aligned (by padding) in case M is not
                   // a good number for loading during backward
                   constexpr decltype(M) kAlignLSE =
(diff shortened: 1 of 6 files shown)


xformers/components/attention/csrc/cuda/mem_eff_attention/attention_forward_generic.cu
```diff
@@ -177,6 +178,18 @@ efficient_attention_forward_generic(
                   p.logsumexp_ptr = compute_logsumexp
                       ? (typename Kernel::lse_scalar_t*)logsumexp.data_ptr()
                       : nullptr;
+                  at::Tensor output_accum;
+                  if (Kernel::kNeedsOutputAccumulatorBuffer) {
```
fmassa, line 182: for my information: this temporary buffer is only needed for K > 128

xformers/components/attention/csrc/cuda/mem_eff_attention/attention_forward_generic.cu
```diff
@@ -143,11 +143,12 @@ efficient_attention_forward_generic(
                       value.stride(1) % Kernel::kAlignmentV == 0,
                       "value is not correctly aligned");
 
-                  res = at::zeros(
+                  res = at::empty(
```
fmassa, line 146: The zeros was not needed before because we handle the kernel differently according if it's the first iteration or not (write vs accumulate)

fmassa (approved):
#shipit!

PR #415, opened 2022-09-22
custommma: Add custom MMA for more optimizations

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #443
* #442
* #440
* #439
* #438
* #434
* #429
* #426
* #421
* __->__ #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* #399
* #398
* #397
* #396
* #395

PR #426, opened 2022-09-23
causalfix: Fix causal after iteration inversion

danthe3rd:
* #443
* #442
* #440
* #439
* #438
* #434
* #429
* __->__ #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* #399
* #398
* #397
* #396
* #395

**SUMMARY**

We now correctly skip the parts of the attention matrix that we shouldn't compute.


fmassa (approved):
Thanks!

For clarity, this fixes a performance regression from the previous commit in the stack, by skipping the elements which are not needed because of the causal part.

PR #429, opened 2022-09-23
prologuemma: Prefetch tensors before mmuls

danthe3rd:
* #443
* #442
* #440
* #439
* #438
* #434
* __->__ #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* #399
* #398
* #397
* #396
* #395

**SUMMARY**

We also try to reuse K for the first matmul between iterations, if it is possible
I also changed how we represent shared-memory to make it easier to change/hack


fmassa (approved):
Thanks!

PR #439, opened 2022-09-27
noqs: Advance pointers to block position

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #443
* #442
* #440
* __->__ #439
* #438
* #434
* #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* #399
* #398
* #397
* #396
* #395

**SUMMARY**

This means we can effectively remove `query_start()` function calls.
This will be useful to support arbitrary sequence lengths

**PERFORMANCE**

<details>
<summary>A100 fw</summary>

```
[----------------------------------------- attention (attn_bias=<class 'NoneType'>) -----------------------------------------]
                                 |  flash[flshatt]  |  37_noqs_1b917ab[cutlass]  |  3_benchmarks_a2c3089[fwd_gen]  |  vanilla 
1 threads: -------------------------------------------------------------------------------------------------------------------
      f16 B=160, M=4096, K=128   |      22646.4     |           12985.0          |              13021.4            |   20676.9
      f16 B=160, M=8192, K=128   |      89871.2     |           51656.2          |              51753.6            |          
      f16 B=256, M=128, K=16     |         29.4     |              23.0          |                 28.5            |      63.3
      f16 B=256, M=128, K=32     |         29.6     |              22.8          |                 27.8            |      62.9
      f16 B=256, M=128, K=64     |         29.5     |              22.8          |                 27.9            |      63.6
      f16 B=256, M=128, K=128    |         49.8     |              37.4          |                 50.5            |      74.8
      f16 B=256, M=512, K=16     |        113.4     |             181.1          |                189.3            |     474.9
      f16 B=256, M=512, K=32     |        131.2     |             184.6          |                196.8            |     489.7
      f16 B=256, M=512, K=64     |        236.5     |             213.9          |                240.8            |     537.0
      f16 B=256, M=512, K=128    |        600.2     |             361.6          |                403.1            |     611.6
      f16 B=256, M=1024, K=16    |        443.0     |             695.3          |                713.3            |    1789.5
      f16 B=256, M=1024, K=32    |        488.0     |             692.1          |                723.0            |    1811.1
      f16 B=256, M=1024, K=64    |        908.5     |             797.4          |                849.8            |    1938.7
      f16 B=256, M=1024, K=128   |       2253.9     |            1361.5          |               1438.6            |    2146.2
[...]

PR #438, opened 2022-09-27
cleanup: Cleanup forward/backward dispatch

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #443
* #442
* #440
* #439
* __->__ #438
* #434
* #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* #399
* #398
* #397
* #396
* #395

**SUMMARY**
Also rename generic->cutlass for kernel name

PR #433, opened 2022-09-23
bw_preproc: Compute delta inside kernels for f16

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* (to be filled)

PR #440, opened 2022-09-27
seqlen: Support variable sequence lengths

danthe3rd:
* #443
* #442
* __->__ #440
* #439
* #438
* #434
* #429
* #426
* #421
* #415
* #412
* #411
* #433
* #432
* #405
* #404
* #403
* #399
* #398
* #397
* #396
* #395

**SUMMARY**

Will make it possible to support Nested Tensors in pytorch.
This is not currently exposed to end-users, as we need to figure
out how we pass the cumulative sequence lengths to the kernels

### DIFF
--- a/xformers/components/attention/csrc/cuda/mem_eff_attention/attention_forward_generic.cu
+++ b/xformers/components/attention/csrc/cuda/mem_eff_attention/attention_forward_generic.cu
@@ -24,7 +24,7 @@
         at::cuda::getDeviceProperties(QUERY.device().index());                \
     const int computeCapability = properties->major * 10 + properties->minor; \
     DISPATCH_BLOCKSIZE(                                                       \
-        VALUE.size(2), ([&]() {                                               \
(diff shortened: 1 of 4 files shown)


xformers/components/attention/csrc/cuda/mem_eff_attention/attention_forward_generic.cu
```diff
@@ -62,19 +62,49 @@
   }
 
 namespace {
+/*
+  There are 2 modes for using this function.
+  (Mode BMK) `num_heads=1` with the heads coalesced into the batch dimension
```
fmassa, line 67: this is not actually a requirement, right? your kernel can take a `[B, L, num_heads, head_dim]` dense tensor and work right?

fmassa (commented):
Thanks!

Can you add some tests to make sure the behavior works as expected?

fmassa (approved):
Thanks!

PR #442, opened 2022-09-28
k_kv: support value.shape[-1] != query.shape[-1]

danthe3rd:
* (to be filled)

**SUMMARY**

Also add a lot of stricter tests

### DIFF
--- a/xformers/ops.py
+++ b/xformers/ops.py
@@ -100,28 +100,37 @@ class AttentionOpBase(torch.autograd.Function):
     SUPPORTED_MAX_K: float
     SUPPORTED_ATTN_BIAS_TYPES: Set[Any] = {type(None)}
     SUPPORTS_DROPOUT: bool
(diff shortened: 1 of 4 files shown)


xformers/ops.py
```diff
@@ -315,8 +300,8 @@ def supports(cls, d: "AttentionOpDispatch") -> bool:
         if sm >= 80:
             matmul_alignment_mn = 4
         if uses_tensorcores:
-            matmul_alignment_mn = max(matmul_alignment_mn, 64 // bits_per_scalar)
-        if d.k % matmul_alignment_mn != 0:
+            matmul_alignment_mn = max(matmul_alignment_mn, 128 // bits_per_scalar)
```
fmassa, line 303: This was a bug in the dispatch before?

fmassa (approved):
Looks great, thanks for the thorough tests Daniel!

xformers/ops.py
danthe3rd: Yes, caught by the new tests

PR #445, opened 2022-09-29
regfix: Fix regression with cu seqlen PR

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #452
* #451
* #449
* #448
* #447
* #446
* __->__ #445
* #444

**PERFORMANCE COMPARISON**

<details>
<summary>A100 fw</summary>

```
[--------------------------------------- attention (attn_bias=<class 'NoneType'>) ---------------------------------------]
                                 |  37_regfix_85c0fb4[cutlass]  |  vanilla  |  flash[flshatt]  |  37_noqs_1b917ab[cutlass]
1 threads: ---------------------------------------------------------------------------------------------------------------
      f16 B=16, M=4096, K=40     |              873.6           |   1942.5  |                  |             882.0        
      f16 B=16, M=16384, K=40    |            12504.9           |  30670.0  |                  |           12638.1        
      f16 B=160, M=4096, K=128   |            12941.5           |  20664.9  |      22646.4     |           12985.0        
      f16 B=160, M=8192, K=128   |            51478.8           |           |      89871.2     |           51656.2        
      f16 B=256, M=128, K=16     |               32.5           |     62.3  |         29.4     |              23.0        
      f16 B=256, M=128, K=32     |               32.9           |     62.1  |         29.6     |              22.8        
      f16 B=256, M=128, K=64     |               33.0           |     61.8  |         29.5     |              22.8        
      f16 B=256, M=128, K=128    |               37.9           |     75.7  |         49.8     |              37.4        
      f16 B=256, M=512, K=16     |              180.0           |    474.8  |        113.4     |             181.1        
      f16 B=256, M=512, K=32     |              184.3           |    489.5  |        131.2     |             184.6        
      f16 B=256, M=512, K=64     |              212.0           |    538.1  |        236.5     |             213.9        
      f16 B=256, M=512, K=128    |              361.1           |    611.1  |        600.2     |             361.6        
      f16 B=256, M=1024, K=16    |              693.3           |   1787.4  |        443.0     |             695.3        
      f16 B=256, M=1024, K=32    |              691.4           |   1818.0  |        488.0     |             692.1        
      f16 B=256, M=1024, K=64    |              790.7           |   1937.9  |        908.5     |             797.4        
      f16 B=256, M=1024, K=128   |             1365.0           |   2147.8  |       2253.9     |            1361.5        
[...]

PR #444, opened 2022-09-29
ci: Enable CI on multiple GPUs

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #452
* #451
* #449
* #448
* #447
* #446
* #445
* __->__ #444

PR #446, opened 2022-09-29
fw_forceinline: Apply forceinline for all device functions

danthe3rd:
* #452
* #451
* #449
* #448
* #447
* __->__ #446
* #445
* #444


fmassa (approved):
LGTM, does it affect performance somehow? I would maybe think that this could add some register pressure, but probably this doesn't matter here?

PR #447, opened 2022-09-30
v100reduc: Implement warp reduction on V100 for same line [5-10% improv]

danthe3rd:
* #452
* #451
* #449
* #448
* __->__ #447
* #446
* #445
* #444


fmassa (approved):
Approving to unblock, but would be good to understand if the mask needs to be changed or not, and if yes why did our tests didn't catch this.

PR #448, opened 2022-09-30
log2ef: Multiply by M_LOG2E in float (rather than double)

danthe3rd:
* #452
* #451
* #449
* __->__ #448
* #447
* #446
* #445
* #444


fmassa (approved):
Thanks!

Does this improve performance somehow?

PR #449, opened 2022-09-30
packed: Support (some) non-contiguous Q,K,V

danthe3rd:
* (to be filled)


xformers/ops.py
```diff
@@ -632,6 +676,7 @@ def memory_efficient_attention(
     Supported formats for inputs/outputs:
         [batch, seqlen, num_heads, K]
         [batch, seqlen, K] (Legacy format)
+    Inputs can be non-contiguous
```
fmassa, line 679: maybe mention that the only constraint is that the stride of the last dimension should be 1?

fmassa (approved):
Looks awesome, thanks a lot!

PR #463, opened 2022-10-06
version number centralization

bottler:
* #466
* __->__ #463

* Autogenerate version.py based on a root directory version.txt like pytorch/vision.
* set a local version based on the git commit when building conda packages


setup.py
```diff
@@ -32,18 +31,21 @@ def fetch_requirements():
...
-        suffix = os.getenv("XFORMERS_VERSION_SUFFIX", "")
-        if version_match:
-            return version_match.group(1) + suffix
-        raise RuntimeError("Unable to find version string.")
+if os.getenv("BUILD_VERSION"):
+    version = os.getenv("BUILD_VERSION")
+else:
+    version_txt = os.path.join(this_dir, "version.txt")
+    with open(version_txt) as f:
+        version = f.readline().strip()
```
fmassa, line 39: Would we want to also add the  git sha in here if we are building from source?
Something like what [torchvision does in here](https://github.com/pytorch/vision/blob/7d2de404372b0a77c5dec825c62f739e75a351ee/setup.py#L42-L43)
This way we are consistent between building with `BUILD_VERSION` (from CI?) and building locally.

Thoughts?

fmassa (approved):
Looks great, thanks Jeremy!

I only have one minor comment that could be addressed in a follow-up PR

fmassa:
nit: also, looks like linter is failing, could you have a look?

PR #471, opened 2022-10-09
update README.md for dev packages

bottler:
* __->__ #471
* #470


bottler:
I think we also need to update the *triton* section. What is the best way to install it? I don't think we should suggest `pip install triton`?

PR #486, opened 2022-10-20
Split xformers.ops

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #518
* #511
* #510
* #509
* #505
* #508
* #504
* #491
* #502
* #490
* #498
* #488
* #487
* __->__ #486

**SUMMARY**
As discussed offline with @fmassa , will make extending new operators clearner

PR #487, opened 2022-10-21
ops.swiglu: Tests, baseline, benchmarks

danthe3rd:
* #518
* #511
* #510
* #509
* #505
* #508
* #504
* #491
* #502
* #490
* #498
* #488
* __->__ #487
* #486

### DIFF
--- a/xformers/ops/swiglu.py
+++ b/xformers/ops/swiglu.py
@@ -1,0 +1,109 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All rights reserved.
+#
+# This source code is licensed under the BSD license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Optional
+
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+
+class SwiGLUFFN_Reference(nn.Module):
+    """
+    Reference implementation of a SwiGLU module
+    """
+
+    def __init__(
+        self,
+        in_features: int,
+        hidden_features: Optional[int] = None,
+        out_features: Optional[int] = None,
+        align_as: int = 8,
+    ) -> None:
+        super().__init__()
+        out_features = out_features or in_features
+        hidden_features = hidden_features or in_features
+        swiglu_hidden_features = int(2 * hidden_features / 3)
+        swiglu_hidden_features = (
+            (swiglu_hidden_features + align_as - 1) // align_as * align_as
+        )
+
+        self.w1 = nn.Linear(in_features, swiglu_hidden_features)
+        self.w2 = nn.Linear(in_features, swiglu_hidden_features)
+        self.w3 = nn.Linear(swiglu_hidden_features, out_features)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x1 = self.w1(x)
+        x2 = self.w2(x)
+        hidden = F.silu(x1) * x2
+        return self.w3(hidden)
+
+    def _ordered_params_for_op(self):
+        """Used for testing - returns ordered arguments for operators"""
+        return [
+            self.w1.weight,
+            self.w1.bias,
+            self.w2.weight,
+            self.w2.bias,
+            self.w3.weight,
+            self.w3.bias,
+        ]
+
+
+def _silu_backward(dy, x):
+    # https://github.com/pytorch/pytorch/blob/563b065f5a4b4055fa6b025c2514b566d5fd9439/aten/src/ATen/native/Activation.cpp#L483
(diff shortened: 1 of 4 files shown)


xformers/ops/swiglu.py
```diff
@@ -0,0 +1,109 @@
...
+
+def _silu_backward(dy, x):
+    # https://github.com/pytorch/pytorch/blob/563b065f5a4b4055fa6b025c2514b566d5fd9439/aten/src/ATen/native/Activation.cpp#L483
+    sigm = 1 / (1 + torch.exp(-x.float()))
+    return (dy.float() * sigm * (1 + x.float() * (1 - sigm))).to(x.dtype)
+
+
+class SwiGLU_Decomposed(torch.autograd.Function):
+    """
+    This is just an example implementation with all
```
fmassa, line 63: It's unclear to me why we would want to expose this in the library -- IMO we should just write reference functional implementations (i.e., not decomposing as `torch.autograd.Function` in the tests and benchmarks.

This is not supposed to be used by the user in any way, right?

xformers/ops/swiglu.py
```diff
@@ -0,0 +1,109 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from typing import Optional
+
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+
+class SwiGLUFFN_Reference(nn.Module):
```
fmassa, line 13: I wonder why provide this implementation? Can't we just test the numerics with a functional API?

fmassa (commented):
Thanks for the PR Daniel!

I have a few comments. In particular, I wouldn't add this `nn.Module` nor the `torch.autograd.Function` in `xformers/ops`, and would instead just implement a vanilla function in tests / benchmarks like the following:
```python
def swiglu(...):
    x1 = F.linear(...)
    x2 = F.linear(...)
    x3 = F.silu(x1)
    x4 = x3 * x3
    return F.linear(...)
```

And for the tests, I would have a wrapper `nn.Module` that just calls into this `swiglu`.

Also, I wouldn't write the backwards by hand myself for the reference implementation, but instead use the autograd implementation as reference. This way we are sure we didn't introduce a bug in our autograd.

From a second look, you might have implemented the backwards by hand with PyTorch to have an idea of the numerics which are expected to differ for the operation, is that right?

Thoughts?

xformers/ops/swiglu.py
```diff
@@ -0,0 +1,127 @@
...
+        dx = dx2 @ w2  # 260us (nn)
+        dw2 = dx2.transpose(-2, -1) @ x  # 245us (nt)
+        db2 = dx2.sum(0)  # 50us
+        dx += dx1 @ w1  # 260us (nn)
+        dw1 = dx1.transpose(-2, -1) @ x  # 245us (nt)
+        db1 = dx1.sum(0)  # 50us
+        return (dx, dw1, db1, dw2, db2, dw3, db3)
+
+
+def functional_swiglu(
```
fmassa, line 111:
```suggestion
def _swiglu_reference(
```
?

xformers/ops/swiglu.py
```diff
@@ -0,0 +1,127 @@
...
+
+        self.w1 = nn.Linear(in_features, swiglu_hidden_features)
+        self.w2 = nn.Linear(in_features, swiglu_hidden_features)
+        self.w3 = nn.Linear(swiglu_hidden_features, out_features)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x1 = self.w1(x)
+        x2 = self.w2(x)
+        hidden = F.silu(x1) * x2
+        return self.w3(hidden)
```
fmassa, line 41: Maybe reuse the function `_swiglu_reference` in here?
```suggestion
        return _swiglu_reference(x, self.w1.weight, self.w1.bias, self.w2.weight, self.w2.bias, self.w3.weight, self.w3.bias
```

xformers/ops/swiglu.py
```diff
@@ -0,0 +1,127 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from typing import Optional
+
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+
+class _SwiGLUModule(nn.Module):
```
fmassa, line 13: If we want to expose this to the user through other ops, we can make this public as well, and expose an `op` argument.
```suggestion
class _SwiGLU(nn.Module):
```

fmassa (approved):
Thanks!

Good to merge, I have a few suggestions for the future

PR #491, opened 2022-10-25
MemEff: Set CUDA stream properly

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #518
* #511
* #510
* #509
* #505
* #508
* #504
* __->__ #491
* #502
* #490
* #498
* #488
* #487
* #486

... and add corresponding tests

PR #492, opened 2022-10-25
MemEff: BUGFIX when a batch has no keys

danthe3rd:
* __->__ #492
* #491


fmassa (approved):
Thanks!

Can you add a test for this case?

danthe3rd:
Causes some performance regression (about 0.5%) so let's not fix it

fmassa:
@danthe3rd what happens then in this case? Does it raise an error or just yield wrong numbers?

If it yields wrong results, then I think this is a silent but that should be fixed.

danthe3rd:
> @danthe3rd what happens then in this case? Does it raise an error or just yield wrong numbers?
> 
> If it yields wrong results, then I think this is a silent but that should be fixed.

Well I'm not even sure what should happen. If there are some queries but no key/no value, I guess it should be zero?
Currently it either outputs NaN or uninitialized values.

PR #496, opened 2022-10-26
Fix CI by updating torchvision

danthe3rd:
We were using torchvision `0.2.2` before, which is fairly old - not sure why.

For some reason, this also updates the clang-format version (I guess?), so I needed to apply it again. Now it matches the output of my local clang-format at least - the difference was only for a single file


fmassa (approved):
Thanks!

I'm approving to unblock, although I suspect this might be a transient issue due to PyTorch releasing 1.13 soon (so maybe some of the binaries haven't been built at the time you checked)

PR #507, opened 2022-11-03
Adding the license title

dblunk88:
Adding the license title, assuming it was forgotten


danthe3rd:
Woops good catch. Let me double check before I merge this

fmassa (approved):
Thanks!

PR #511, opened 2022-11-04
Include git hash and install date in version

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #518
* __->__ #511
* #510
* #509
* #505
* #508
* #504
* #491
* #502
* #490
* #498
* #488
* #487
* #486

PR #509, opened 2022-11-04
Raise if kernel not supported on device

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #518
* #511
* #510
* __->__ #509
* #505
* #508
* #504
* #491
* #502
* #490
* #498
* #488
* #487
* #486

PR #505, opened 2022-11-03
Benchmarks for entire models

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #518
* #511
* #510
* #509
* __->__ #505
* #508
* #504
* #491
* #502
* #490
* #498
* #488
* #487
* #486

PR #504, opened 2022-11-02
SwiGLU: Support for no bias

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #518
* #511
* #510
* #509
* #505
* #508
* __->__ #504
* #491
* #502
* #490
* #498
* #488
* #487
* #486

**PERFORMANCE**

```
[-------------------------- swiglu_fw --------------------------]
                                          |  optimized  |  eager 
1 threads: ------------------------------------------------------
      b16    B=9456, I=1536, H=4096 bias  |    1338.5   |  1568.4
      b16    B=9456, I=1536, H=4096 nobi  |    1268.3   |  1477.1
      f16    B=9456, I=1536, H=4096 bias  |    1305.3   |  1607.4
      f16    B=9456, I=1536, H=4096 nobi  |    1291.7   |  1493.0
      f16.ac B=9456, I=1536, H=4096 bias  |    1445.1   |  1754.7
      f16.ac B=9456, I=1536, H=4096 nobi  |    1434.0   |  1640.5
      b16    B=4440, I=1536, H=4096 bias  |     580.7   |   726.5
      b16    B=4440, I=1536, H=4096 nobi  |     578.7   |   725.1
      f16    B=4440, I=1536, H=4096 bias  |     597.4   |   736.5
      f16    B=4440, I=1536, H=4096 nobi  |     598.0   |   732.3
      f16.ac B=4440, I=1536, H=4096 bias  |     712.8   |   853.8
      f16.ac B=4440, I=1536, H=4096 nobi  |     701.0   |   841.7
      b16    B=4728, I=1536, H=4096 bias  |     620.3   |   772.6
      b16    B=4728, I=1536, H=4096 nobi  |     618.2   |   744.7
      f16    B=4728, I=1536, H=4096 bias  |     634.9   |   785.9
      f16    B=4728, I=1536, H=4096 nobi  |     633.7   |   750.8
      f16.ac B=4728, I=1536, H=4096 bias  |     746.8   |   897.1
      f16.ac B=4728, I=1536, H=4096 nobi  |     740.0   |   854.1
      b16    B=4728, I=1536, H=1024 bias  |     160.1   |   199.3
      b16    B=4728, I=1536, H=1024 nobi  |     158.6   |   196.4
      f16    B=4728, I=1536, H=1024 bias  |     163.6   |   202.1
      f16    B=4728, I=1536, H=1024 nobi  |     161.9   |   198.8
[...]

PR #518, opened 2022-11-10
Address nits from previous PR stack + bump version

danthe3rd:
* __->__ #518
* #511
* #510
* #509
* #505
* #508
* #504
* #491
* #502
* #490
* #498
* #488
* #487
* #486


version.txt
```diff
@@ -1 +1 @@
-0.0.14.dev
+0.0.15.dev
```
fmassa, line 1: I would have preferred if we bumped the version in a follow-up PR

setup.py
```diff
@@ -133,8 +133,6 @@ def get_flash_attention_extensions(cuda_version: int, extra_compile_args):
                 "nvcc": extra_compile_args.get("nvcc", [])
                 + [
                     "-O3",
-                    "-U__CUDA_NO_HALF_OPERATORS__",
-                    "-U__CUDA_NO_HALF_CONVERSIONS__",
```
fmassa, line 137: So this wasn't needed in the end?

fmassa (approved):
LGTM, thanks!

Also, if we are bumping the version, we might need to bump it as well on pypi. cc @dianaml0

setup.py
danthe3rd: Nope, I just copy pasted it because it was in cutlass's build process I think. Good catch

PR #547, opened 2022-11-28
Wheels: One artefact file per config

danthe3rd:
Also makes it run every week

cc @AbdBarho let me know if this looks good to you

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -2,7 +2,6 @@ name: Build wheels
 
 on:
   workflow_dispatch: {}
-  pull_request: {}
   schedule:
     # Run at 3am every monday
     - cron: '0 3 * * 1'
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -2,6 +2,7 @@ name: Build wheels
 
 on:
   workflow_dispatch: {}
+  pull_request: {}
   schedule:
     # Run at 3am every monday
     - cron: '0 3 * * 1'
@@ -122,7 +123,7 @@ jobs:
 
       - name: Install build dependencies
         run: >
-          $PY -m pip install wheel setuptools torch==${{ matrix.config.torch_version }} -r requirements.txt --extra-index-url ${{ matrix.config.pip_index }} &&
+          $PY -m pip install wheel setuptools -r requirements.txt --extra-index-url ${{ matrix.config.pip_index }} &&
           git config --global --add safe.directory "*"
 
       - name: Build wheel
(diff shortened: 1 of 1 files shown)


AbdBarho:
@danthe3rd look good! glad to see it the wheels being put to good use.

.github/workflows/wheels.yml
```diff
@@ -108,16 +114,22 @@ jobs:
...
 
+      - name: Setup proper pytorch dependency in "requirements.txt"
+        run: >
+          sed -i '/pytorch/d' ./requirements.txt
+          echo "pytorch == ${{ matrix.config.torch_version_constraint }}" >> ./requirements.txt
+          cat ./requirements.txt
+
       - name: Install build dependencies
         run: >
           $PY -m pip install wheel setuptools torch==${{ matrix.config.torch_version }} -r requirements.txt --extra-index-url ${{ matrix.config.pip_index }} &&
```
AbdBarho, line 125: I wonder if `torch==${{ matrix.config.torch_version }}` is still necessary after doing the new update, would it make sense to just write the version with cuda in the requirements file and remove the torch version from here?

danthe3rd: Good point - will update

.github/workflows/wheels.yml
```diff
@@ -33,17 +37,20 @@ jobs:
...
 
           - torch_version: "1.13.0+cu116"
+            torch_version_constraint: "1.13"
             cuda_version: "11.6.2"
             pip_index: https://download.pytorch.org/whl/cu116
             cuda_run_file: https://developer.download.nvidia.com/compute/cuda/11.6.2/local_installers/cuda_11.6.2_510.47.03_linux.run
 
           # torch 1.12 is built with cuda 11.3 and 11.6, we ignore 11.3 for now
           - torch_version: "1.12.1+cu116"
+            torch_version_constraint: "1.12"
```
fmassa, line 53: nit: is this constraint in line with what is being used by `torch_version` just above? (I see `1.12.1`).

Also, wouldn't it be a bit more robust to make the `torch_version` be a function of `torch_version_constraint` + `cuda_version` ?

.github/workflows/wheels.yml
```diff
@@ -108,16 +115,22 @@ jobs:
           submodules: recursive
           path: "."
 
+      - name: Setup proper pytorch dependency in "requirements.txt"
+        run: >
+          sed -i '/pytorch/d' ./requirements.txt
+          echo "pytorch == ${{ matrix.config.torch_version_constraint }}" >> ./requirements.txt
+          cat ./requirements.txt
```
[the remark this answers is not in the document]
danthe3rd: Fixed it - thanks :)

PR #548, opened 2022-11-28
FlashAttention needs contiguous gradients

fmassa:
This was not visible before because the test was creating tensors of the wrong dtype, which when casted to the right dtype would yield contiguous tensors


danthe3rd (approved):
Thanks a lot for spotting!

PR #539, opened 2022-11-24
Improve build time by ~30%

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #467
* #549
* #540
* __->__ #539

... by reducing the number of ATen imports, and skipping them altogether when building the actual kernels

13mn -> 9mn on Sm61 (CI, does not build flash)

PR #467, opened 2022-10-06
MemEff: Accumulate in f32 for bw

danthe3rd:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #467

**PERFORMANCE**

This makes performance worse in f16 :(
But I think we need it for stability

<details>
<summary>bw P100/V100 (f32/f16)</summary>

```
[---------------------------------------- attention backward (attn_bias=<class 'NoneType'>) ----------------------------------------]                                                    
                                                         |  48_accf32_6f8e2f15  |  56_base_02bf6b4e  |  vanilla   |  57_tmpT_b516aec4
1 threads: --------------------------------------------------------------------------------------------------------------------------
  (Quadro_GP100)          f16 B=384, M=197, H=1, K=88    |         6289.6       |        6936.0      |    2183.6  |        6940.0    
                          f32 B=384, M=197, H=1, K=88    |         8793.3       |        9446.6      |    2175.1  |        9429.2    
                          f16 B=384, M=197, H=1, K=80    |         5989.9       |        6596.2      |    2146.8  |        6608.6    
                          f32 B=384, M=197, H=1, K=80    |         8427.1       |        8993.1      |    2134.0  |        9030.5    
                          f16 B=384, M=197, H=1, K=64    |         3347.0       |        3527.4      |    1799.3  |        3538.3    
                          f32 B=384, M=197, H=1, K=64    |         5563.1       |        5984.2      |    1801.6  |        5980.9    
                          f16 B=1024, M=197, H=1, K=88   |        15680.5       |       17424.8      |    5671.9  |       17452.6    
                          f32 B=1024, M=197, H=1, K=88   |        23784.2       |       25542.8      |    5664.0  |       25578.6    
                          f16 B=1024, M=197, H=1, K=80   |        14935.5       |       16581.3      |    5559.5  |       16587.3    
                          f32 B=1024, M=197, H=1, K=80   |        22767.6       |       24354.1      |    5550.7  |       24362.0    
                          f16 B=1024, M=197, H=1, K=64   |         8331.1       |        8671.5      |    4644.9  |        8695.3    
                          f32 B=1024, M=197, H=1, K=64   |        15061.6       |       16148.8      |    4650.0  |       16201.4    
                          f16 B=512, M=197, H=1, K=80    |         7594.7       |        8306.6      |    2824.6  |        8336.5    
[...]

PR #549, opened 2022-11-29
Reduce binary size - no debug info for binaries

danthe3rd:
* #467
* __->__ #549
* #540
* #539


fmassa (approved):
Looks great, thanks!

PR #551, opened 2022-11-30
[CI] Ensure we don't break windows build

danthe3rd:


fmassa (approved):
LGTM, thanks!

At some point we might consider using the new tooling from PyTorch for more easily generating / building jobs.

Here are some examples: 
For building wheels: https://github.com/pytorch/vision/blob/main/.github/workflows/build-wheels-linux.yml

For testing: https://github.com/pytorch/vision/blob/main/.github/workflows/test-linux-gpu.yml

PR #554, opened 2022-12-06
[minor] Fix tracing a mem efficient attention enabled part

blefaudeux:
## What does this PR do?
- Max_k is an int really (well, a uint)
- a float there results in ` RuntimeError: value cannot be converted to type int64_t without overflow` when tracing a model

Not that I'm really far from an expert there, so any other take welcome. I also know that TorchScript may not be here to stay, but it felt like a small enough fix

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/xformers/ops/memory_efficient_attention.py
+++ b/xformers/ops/memory_efficient_attention.py
@@ -104,7 +104,7 @@ class AttentionOpBase(torch.autograd.Function):
     }
     SUPPORTED_DEVICES: Set[str]
     SUPPORTED_DTYPES: Set[torch.dtype]
-    SUPPORTED_MAX_K: float
+    SUPPORTED_MAX_K: int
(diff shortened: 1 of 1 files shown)


danthe3rd:
The reason we are using a float is because it supports "inf" to mean no limit. This should be updated in some way as well I guess?

blefaudeux:
> The reason we are using a float is because it supports "inf" to mean no limit. This should be updated in some way as well I guess?

ahh, I missed that. Squaring a circle, isn't this a little theoretical ? (2 ** 63 would be very close to inf for mere mortals, no ?)

danthe3rd:
Also we're in the middle of refactoring these attention classes internals, so maybe you can contribute a test and we can take care of the fix?

blefaudeux:
sorry I completely forgot about this PR, fixed since, removing the noise here..

PR #578, opened 2022-12-12
fix cutlass SUPPORTED_MAX_K

zzpmiracle:
fix op.support bug
image


danthe3rd:
cc @blefaudeux

fmassa:
@danthe3rd yes, I think it's safe to assume that K > 65000 is not going to happen.

fmassa (approved):
Thanks!

As @danthe3rd mentioned, if torchscriptability is your end goal, would you maybe send another PR adding tests for torchscriptability, so that we don't break it in the future?

PR #579, opened 2022-12-12
Move all C++/CUDA source files into csrc

danthe3rd:
Should make it easier to share C++ code between components

**TEST PLAN**

CI + ran SwiGLU manually on A100 (as we don't run SwiGLU on CI):

```bash
python -m pytest /scratch/dhaziza/xformers/tests/test_swiglu.py
```


fmassa (commented):
This will break some of our internal tests. Could you look into it?

Thanks!

danthe3rd:
Will be done in https://github.com/fairinternal/xformers/pull/405

PR #584, opened 2022-12-13
Attempt to fix CI

danthe3rd:
Weird errors about missing .so file that I can't repro when creating clean envs ...


danthe3rd:
Does not seem needed

PR #588, opened 2022-12-13
Publish Wheels to PyPi

AbdBarho:
Related #533 #573 #532  #534 #523 #473 (damn, this is getting out of hand).

A few decisions here:
* We only upload the wheels built against the latest torch version (1.13) with latest cuda (11.7)
* the combination 1.13 and 11.6 was removed with the hope that the other version compensates.
* The 1.12 version is left in for users who still haven't updated yet, but it is not uploaded to pypi.
* the version string is as discussed in https://github.com/facebookresearch/xformers/issues/533#issuecomment-1348736095
  * _release candidate_ with a suffix `dev` (not `rc`) to stay consistent with conda
  * we should probably consolidate the version generation to one place

Example run:
https://github.com/AbdBarho/xformers-wheels/actions/runs/3689261975

Published Wheels:
https://test.pypi.org/project/formers/0.0.15.dev376/#files

TODO (for future PRs):
* Publish on tagged commits
* Create source distribution

PR #590, opened 2022-12-14
Upload source distribution to PyPi

AbdBarho:
## What does this PR do?

Upload source distribution to PyPi. Continuation of #588 

I am unsure if this is strictly necessary since we basically upload wheels for every supported platform, but it won't hurt `¯\_(ツ)_/¯`

Related #533 

I have considered consoldating  the `env` from the last 2 steps to the global env, but I don't want any steps (especially external actions) to access the token.

Example sdist:
https://test.pypi.org/project/formers/0.0.16rc379/#files

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -43,6 +43,11 @@ jobs:
             pip_index: https://download.pytorch.org/whl/cu116
             cuda_run_file: https://developer.download.nvidia.com/compute/cuda/11.6.2/local_installers/cuda_11.6.2_510.47.03_linux.run
             publish: false
+        include:
+          # publish source distribution only from this runner
+          - os: ubuntu-22.04
+            python: "3.10"
+            sdist: true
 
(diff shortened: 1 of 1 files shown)


danthe3rd:
This looks great - thanks a lot! Is it ready to merge?
Btw, we are limited to 10GB of storage on pypi. One version takes ~300MB, so we can "afford" ~30 versions.
I suggest we keep the last ~10 RC, and last ~20 tagged versions - I'm working on a script to automate the cleaning of the previous RC.

AbdBarho:
it is ready to merge!

The answer to your question depends entirely on how often do you plan on releasing new versions. My biggest concern is old versions of the library not being available anymore.

~20 tagged versions + release once a month = ~1.5 years average lifetime of a version, which doesn't sound like much.

Maybe after 1.5 years the project is big enough to request another storage increase from pypi :p

This is totally up to you and maybe others on the team to decide, I don't really mind either way.

danthe3rd:
Yeah, our versions will also quickly be outdated, as they rely on a specific PyTorch version. We could keep the source distribution tho for tagged releases

PR #591, opened 2022-12-15
Publish to PyPi on Git Tag

AbdBarho:
The final piece in the pip trilogy #533.

Continuation of #588 and #590.

This should hopefully be all the work needed to have pip wheels always up-to-date.

I added a check to see if the version in `version.txt` matches the git tag, not sure if necessary, maybe it is okay if they are different?

After this, I intend to do a clean up of the workflow code since it is starting to get a bit messy, especially with all the os specific stuff.

Also, there was a mention of having scripts that cleans up old releases from PyPi, @danthe3rd ping me if needed.


.github/workflows/wheels.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - "*"
```
danthe3rd, line 8: Can we have a sort of regexp here to ensure we only match tags that look like a version number?

danthe3rd (approved):
Thanks a lot! This looks great - just a small nit on tags that should trigger this, and we can finalize the trilogy :)

.github/workflows/wheels.yml
AbdBarho: Assuming semantic versioning, so something like `[0-9]+.[0-9]+.[0-9]+` ?

danthe3rd: We don't need the "v" prefix

danthe3rd:
I'll merge once the CPU tests pass :)

danthe3rd:
Awesome! Thanks a lot - a lot of users will be thankful that we finally can have binary wheels - and also with windows supported :)
I'll see internally when we are ready to bump to 0.0.16 - hopefully soon

PR #592, opened 2022-12-15
MHA: Stricter input validation

danthe3rd:
* #594
* #593
* __->__ #592

### DIFF
--- a/xformers/ops/fmha/__init__.py
+++ b/xformers/ops/fmha/__init__.py
@@ -307,6 +307,7 @@ def _memory_efficient_attention_forward(
             f"xformers.memory_efficient_attention: Operator {op.NAME} does not support this input"
         )
 
+    inp.validate_bmhk()
     out, *_ = op.apply(inp, needs_gradient=False)
     return out.reshape(output_shape)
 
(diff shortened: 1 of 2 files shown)


xformers/ops/fmha/common.py
```diff
@@ -99,6 +99,43 @@ def normalize_bmhk(self) -> Tuple[int, ...]:
             self.value = self.value.unsqueeze(2)
         return output_shape
 
+    def validate_bmhk(self) -> None:
+        qkv = (self.query, self.key, self.value)
+        if tuple(x.ndim for x in qkv) != (4, 4, 4):
+            raise ValueError(
+                f"Query/Key/Value should have BMHK format.\n"
```
fmassa, line 106: This error could be misleading. The user can pass a `BMK` tensor as well, and will be confused as to why this mentions that only 4d inputs are valid

xformers/ops/fmha/__init__.py
```diff
@@ -321,6 +322,7 @@ def _memory_efficient_attention_forward_requires_grad(
         raise ValueError(
             f"xformers.memory_efficient_attention: Operator {op.NAME} does not support this input"
         )
+    inp.validate_bmhk()
```
fmassa, line 325: I would have seen those checks right in the beginning of the function, just before `normalize_bmhk`, this way we could also capture the fact that the tensors can be 3d or 4d

xformers/ops/fmha/common.py
```diff
@@ -99,6 +99,43 @@ def normalize_bmhk(self) -> Tuple[int, ...]:
             self.value = self.value.unsqueeze(2)
         return output_shape
 
+    def validate_bmhk(self) -> None:
```
fmassa, line 102: if you move this to the beginning of the function, I would maybe call it `validate_inputs`?

PR #597, opened 2022-12-15
Support v prefix for tags

AbdBarho:
## What does this PR do?

Support using `v` as prefix for a tagged version.

Fix RC release error.

The cpu test fails because of something unrelated to this PR.


danthe3rd:
yes the test failing is unrelated. Thanks a lot for this contribution! Merging :)

PR #598, opened 2022-12-15
Force CUDA when TORCH_CUDA_ARCH_LIST is set

danthe3rd:
* __->__ #598


fmassa (approved):
LGTM, thanks!

BTW, it might be worth seeing if we might want to propose those changes to torchvision as well.

PR #599, opened 2022-12-15
TensorWithSeqLen.to_tensor_list() returns torch.Tensors

danthe3rd:


fmassa (approved):
Thanks for fixing this!

PR #600, opened 2022-12-16
Consolidate Windows Runners' Setup

AbdBarho:
## What does this PR do?

Reduces code duplication, as a continuation of #591

After this, I will investigate possibilities of speeding up the build, not sure If I can squeeze something out...


danthe3rd (approved):
Nice refactor! Thanks a lot!

PR #602, opened 2022-12-18
Relax pytorch dependency version

AbdBarho:
## What does this PR do?

Fixes an issue introduced by #590

we pin pytorch's version while generating the wheels, and since we re-use the runner to generate the source distribution; the pinned version gets carried over.

This PR unpins the version.

Maybe it would make sense to use a different runner for the source distribution, but this change will be required nonetheless, since pytorch is pinned per default in requirements.txt.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -33,15 +33,15 @@ jobs:
           - "3.9"
           - "3.10"
         config:
-          - torch_version: "1.13.0+cu117"
-            torch_version_constraint: "1.13"
+          - torch_version: "1.13.1+cu117"
+            torch_version_constraint: "1.13.1"
             cuda_version: "11.7.1"
             pip_index: https://download.pytorch.org/whl/cu117
             cuda_run_file: https://developer.download.nvidia.com/compute/cuda/11.7.1/local_installers/cuda_11.7.1_515.65.01_linux.run
             publish: true # publish only latest version
 
           - torch_version: "1.12.1+cu116"
-            torch_version_constraint: "1.12"
+            torch_version_constraint: "1.12.1"
             cuda_version: "11.6.2"
             pip_index: https://download.pytorch.org/whl/cu116
             cuda_run_file: https://developer.download.nvidia.com/compute/cuda/11.6.2/local_installers/cuda_11.6.2_510.47.03_linux.run
@@ -137,7 +137,7 @@ jobs:
 
       - uses: actions/upload-artifact@v3
         with:
-          name: ${{ matrix.os }}-py${{ matrix.python }}-torch${{ matrix.config.torch_version }}.zip
+          name: ${{ matrix.os }}-py${{ matrix.python }}-torch${{ matrix.config.torch_version }}
           path: dist/*.whl
 
       - name: Upload wheel to PyPi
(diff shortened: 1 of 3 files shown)


danthe3rd:
This makes sense, although I'm not sure this is possible with pypi. Can different wheels for the same package version have different requirements?

danthe3rd:
I mean, what happens when you try to pip-install with the wrong pytorch version? Does it upgrade pytorch? Or does it build from source?

AbdBarho:
@danthe3rd 
I think the wheels are preferred and prioritized, unless there are no wheels for the given platform. I do believe we have wheels for all (usual) platforms so install and building the source distribution is highly unlikely.

However, I have noticed  a different problem, which is that the requirement `torch == 1.x` always resolves to `1.x.0`, which is not what we want.

You can even see this in the workflow logs:
https://github.com/facebookresearch/xformers/actions/runs/3740207261/jobs/6348305710#step:8:28

I have pushed a fix, we can leave the topic of the source distribution open for now.

danthe3rd:
> I have pushed a fix, we can leave the topic of pinning version of source distribution as is for now.

I discussed with @bottler and he does not think this is an issue, so would be happy to merge the PR to unpin pytorch's version on the source distribution.
Btw, I'm also wondering if the source distribution is correct (I don't think we ever tested it). We should make sure it includes the flash-attention/cutlass source code (in third_party)

> the wheels should be compatible with all patch versions of a pytorch's minor release.

I don't think it's the case (cc @fmassa ) as the C++ ABI can change between minor releases, which makes the compiled library incompatible ...

AbdBarho:
@danthe3rd 

> would be happy to merge the PR to unpin pytorch's version on the source distribution.

Ok I can re-unpin torch, would it make sense to revert it back to `>=1.12` as it is currently in the requirements.txt? or leave it completely open?

> I don't think it's the case as the C++ ABI can change between minor releases, which makes the compiled library incompatible ...

Did you mean between patch releases? I expect change between minor releases `1.12` and `1.13`, but is it also the case for patch releases `1.13.0` and `1.13.1`?

danthe3rd:
> would it make sense to revert it back to >=1.12 as it is currently in the requirements.txt? or leave it completely open?

It makes sense to fallback to the version in `requirements.txt` yes :)

> is it also the case for patch releases 1.13.0 and 1.13.1?

That's unfortunately the case yes. it might work "by accident", but we can't expect this to be true. The python/C++ API is stable, but ABI will always break (so we need to recompile)

danthe3rd:
(also btw, we should probably build for 1.13.1 now, as it's been out for a week)

AbdBarho:
@danthe3rd I managed to get the source distribution running (at least on my machine), however, I think the code changes I made in `setup.py` could be cleaner... If you have any ideas that could be great.

Also, in `MANIFEST.in`, I added what I saw necessary from `setup.py`, which is maybe too much or not enough...

The source dist is 1.4MB, which is bigger than I expected.

You can try a source build from here: 
https://test.pypi.org/project/formers/0.0.16rc391/#files

danthe3rd (approved):
Thanks a lot! This looks great. I'll merge once CI is green

The code you modified in `setup.py` was already a hack before, now it at least it will work - so I'm fine with that :)
As a follow-up, we would a way to test that source-distr can still build at some point, so we don't break it in the future.

PR #610, opened 2022-12-20
conda builds for releases

bottler:
* __->__ #610


danthe3rd (approved):
This looks good to me! Before merging, let's try to refactor to use [the same code](https://github.com/facebookresearch/xformers/blob/main/packaging/compute_wheel_version.py) used for the pip wheels as well.

cc @AbdBarho

bottler:
Updated to unify the code for determining whether we are doing a proper release. However, the version number when we aren't is inconsistent. Conda uses `f"{code_version}.dev{num_commits}+git.{git_hash}"` but wheel uses `f"{version}rc{num_commits}`. Can we make wheel like conda, since it is newer? Perhaps not every commit should be considered a release candidate.

danthe3rd:
> Conda uses `f"{code_version}.dev{num_commits}+git.{git_hash}"` but wheel uses `f"{version}rc{num_commits}`. Can we make wheel like conda, since it is newer?

It would be great to have something similar in both indeed.

We have constraints in pypi, which made us use "rc" releases there:
(1) local suffixes are not supported (eg +gitXXX)
(2) we don't want people to install `.dev` versions by mistake - so we use `rc`. `rc` versions are only installed if you specify the `--pre` argument to `pip` (so it needs to be explicit)

bottler:
I think without `--pre` you won't get dev versions either, so (2) should not matter. So I *could* change wheels to f"{code_version}.dev{num_commits}". But I think it would break the chronology - `pip install --pre xformers` might prefer `0.0.16rc392` to `0.0.16.dev393`, based on the line *Developmental releases are ordered by their numerical component, immediately before the corresponding release (and before any pre-releases with the same release segment)"* in [pep440](https://peps.python.org/pep-0440/#developmental-releases). How about doing nothing on this now and maybe changing to `dev` after the next release?

Having a difference in that conda packages have the local part and pip ones don't is fine by me. The hash is slightly useful.

danthe3rd:
Maybe it makes more sense to move the conda releases to use `rc` if it's possible, to respect the ordering?

> But I think it would break the chronology - pip install --pre xformers might prefer 0.0.16rc392 to 0.0.16.dev393

Also we will eventually remove older packages, so the "rc" ones will eventually disappear

> Having a difference in that conda packages have the local part and pip ones don't is fine by me. The hash is slightly useful.

Agreed

PR #619, opened 2023-01-05
conda build for PyTorch 1.13.1

bottler:
* __->__ #619


danthe3rd (approved):
Thanks a lot!
Btw, can't we build for CUDA 11.7 only? Isn't there some compatibility between minor releases of CUDA supposedly? It seems to work in practice but I'm not sure about the side-effects

bottler:
xlformers could probably rely on the python stable ABI. PyTorch maybe as well.

PR #638, opened 2023-01-12
fairinternal conda => artifacts

bottler:
* __->__ #638


bottler:
> Thanks @bottler ! This adds some complexity to this script - we could move this complexity back into the CI script rather I believe, but that can be addressed later

That was deliberate. I argue it's easier when logic is in python than in bash.

Issue #640, opened 2023-01-12
Proposal for separate passing of causal flag and other types of bias


jfc4050:
# 🚀 Feature

there's performance advantages to treating causal masking as a special case since it can be used by the kernel as a signal to skip computation. This is already supported by the CUTLASS implementation, its just not exposed in the python API

## Motivation

here's some benchmarks showing additive bias and causal being used simultaneously, generally about 2x faster when causal is enabled

```
[----------------------- attention (attn_bias=<class 'torch.Tensor'>) -----------------------]
                                                                        |  optimized  |  eager
1 threads: -----------------------------------------------------------------------------------
      f16 B=4, M=4096, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=False  |     38.2    |   73.2
      b16 B=4, M=4096, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=False  |     38.8    |   73.5
      f16 B=4, M=4096, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=True   |     19.7    |   73.1
      b16 B=4, M=4096, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=True   |     20.1    |   73.6
      f16 B=8, M=4096, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=False  |     76.2    |
      b16 B=8, M=4096, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=False  |     77.6    |
      f16 B=8, M=4096, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=True   |     39.3    |
      b16 B=8, M=4096, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=True   |     39.9    |
      f16 B=4, M=8192, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=False  |    151.9    |
      b16 B=4, M=8192, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=False  |    154.5    |
      f16 B=4, M=8192, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=True   |     77.3    |
      b16 B=4, M=8192, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=True   |     78.6    |
      f16 B=8, M=8192, H=64, K=128, p=0.3,  BiasT=Tensor, Causal=False  |    303.4    |
[...]

PR #643, opened 2023-01-13
RFC: AttentionBias v2

danthe3rd:
See also https://github.com/facebookresearch/xformers/issues/640

Adds support for combinations of different sorts of biases:
- Causal
- Bias (coming with https://github.com/facebookresearch/xformers/pull/587)
- Block-diagonal (used for different seqlen per batch element)

We need to rename "LowerTriangularMask" because when added with a block-diagonal mask it's no longer causal:

```
# A (block-diagonal for q_len=[2, 2] / k_len=[3, 2])
0 0 0 * *
0 0 0 * *
* * * 0 0
* * * 0 0

# B (lower triangular)
0 * * * *
0 0 * * *
0 0 0 * *
0 0 0 0 *

# A + B 
0 * * * *
0 0 * * *
* * * * *
* * * 0 *

# A + causal (what most ppl want)
0 * * * *
0 0 * * *
* * * 0 *
* * * 0 0
```

cc @jfc4050


jfc4050:
this works for our purposes!

danthe3rd:
Merged in https://github.com/facebookresearch/xformers/commit/d23d04e71a5776cc47b58b8d92cdad9bfb41a36f

PR #838, opened 2023-08-31
support MQA/GQA in flash_attn

zzpmiracle:


danthe3rd:
Hi,

Thanks for opening this PR! Indeed at the moment we don't support GQA. We're not 100% clear on what is the best way to support it through the API - and in a way that supports most of the operators, AND the bw pass.
For the backward pass, ideally we don't want to have to deal with inputs broadcasting - otherwise we need to do a reduction in the BW pass.
I'll circle back with the team and try to come up with a decision next week

danthe3rd:
> I think maybe we could support forward first.

Yes totally. But I'm worried that for BW support we will need to change the API in the FW as well. And it would be better to avoid breaking backward compatibility. I'll get back to you as soon as possible :)

danthe3rd:
Some update here, we have a fix working. The idea is to accept 5-dim inputs (BMGHK) where G corresponds to the number of head groups.
We will ship that hopefully soon, we have a few internal tests broken that we need to address first however

CanyonWind:
Hi! following up here -- any plan for supporting backward pass? We'd need to finetune with MQA to gain the best quality, otherwise, inference-only changes will harm the quality. Many thanks

PR #916, opened 2023-11-06
Decouple the kernel_forward.h/kernel_backward.h from PyTorch Dependencies

Li-Jiajie:
## What does this PR do?
Currently, the Attention Dropout feature relies on some PyTorch functions, and the HAS_PYTORCH macro is used in the kernel_forward.h/kernel_backward.h files to indicate whether there is a PyTorch environment. If not, the Dropout feature cannot be used.
At present, the ***generic.cu files generate seed and offset using PhiloxCudaState, while the kernel_forward.h/kernel_backward.h obtain seed and offset through the unpack function.
Perhaps we can directly place the offset and seed in the Kernel::Params parameters to avoid the dependency on PyTorch in the kernel_forward.h/kernel_backward.h.

Background:
I would like to port the Attention implementation from XFormers to a framework other than PyTorch. Based on this PR, I only need to adapt to other frameworks according to the attention_*_generic.cu files, as the kernel_forward.h and kernel_backward.h files are framework-agnostic and no further modifications to other files are required.

## Before submitting

- [ ] Did you have fun?
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -246,7 +246,6 @@ def get_extensions():
         sources += source_cuda
         include_dirs += [sputnik_dir, cutlass_dir, cutlass_examples_dir]
         nvcc_flags = [
-            "-DHAS_PYTORCH",
             "--use_fast_math",
             "-U__CUDA_NO_HALF_OPERATORS__",
             "-U__CUDA_NO_HALF_CONVERSIONS__",
--- a/xformers/csrc/attention/cuda/fmha/attention_backward_generic.cu
+++ b/xformers/csrc/attention/cuda/fmha/attention_backward_generic.cu
@@ -309,7 +309,6 @@ mem_efficient_attention_backward_cutlass(
     }
 
     if (use_dropout) {
-      p.rng_engine_inputs = rng_engine_inputs;
       p.dropout_prob = dropout_p;
       p.dropout_philox_seed = rng_seed;
       p.dropout_philox_offset = rng_offset;
--- a/xformers/csrc/attention/cuda/fmha/attention_backward_generic.cu
+++ b/xformers/csrc/attention/cuda/fmha/attention_backward_generic.cu
@@ -152,7 +152,6 @@ mem_efficient_attention_backward_cutlass(
   at::Tensor workspace;
 
   const bool use_dropout = std::fpclassify(dropout_p) != FP_ZERO;
-  at::PhiloxCudaState rng_engine_inputs(rng_seed, rng_offset);
 
   cudaDeviceProp* p = at::cuda::getDeviceProperties(query.device().index());
   const int computeCapability = p->major * 10 + p->minor;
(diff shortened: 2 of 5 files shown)


danthe3rd (commented):
Hi @Li-Jiajie 
Thanks for opening this PR! First I'm curious to know the name of the framework you want to port this to (if you can shared).
Regarding the content of the PR, I believe this is a good idea (plus also will make compilation a bit faster because pytorch users don't have to import aten headers!).
However, I'm a bit worried about how this will work with CUDA Graphs. There is special "trick" in `at::cuda::philox::unpack` to allow [to replay cuda graphs with a different random seed ](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cuda/detail/UnpackRaw.cuh#L16), where the seed used is actually a pointers to a GPU memory address (so the seed can change between graph replays).
The changes you bring would break this behavior. I'm not sure what's the best way to move forward there - maybe @fmassa @drisspg @cpuhrsch you have an idea?

Li-Jiajie:
> Hi @Li-Jiajie Thanks for opening this PR! First I'm curious to know the name of the framework you want to port this to (if you can share). Regarding the content of the PR, I believe this is a good idea (plus also will make compilation a bit faster because pytorch users don't have to import aten headers!). However, I'm a bit worried about how this will work with CUDA Graphs. There is special "trick" in to allow [to replay cuda graphs with a different random seed ](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cuda/detail/UnpackRaw.cuh#L16), where the seed used is actually a pointers to a GPU memory address (so the seed can change between graph replays). The changes you bring would break this behavior. I'm not sure what's the best way to move forward there - maybe @fmassa @drisspg @cpuhrsch you have an idea?`at::cuda::philox::unpack`

I use TensorFlow and want to implement a custom Attention OP to try accelerate the Attention part of my model.
I general understand the intention of PhiloxCudaState to save pointers in the capture state. So I think perhaps it can be modified to: when passing to kernel::Params, convert the offset and seed to actual values based on the capture_ flag?

But I still have some doubts. In the forward section, `rng_engine_inputs = gen->philox_cuda_state(B * num_heads * M * N); ` generates PhiloxCudaState, which may be a pointer (capture_==true) or a value. At the end of the forward, we directly memcpy the union space of seed and offset and return it as the input for backward. That is to say, in backward, we don't know whether the seed and offset are addresses or values?
In backward, we construct the PhiloxCudaState using `at::PhiloxCudaState rng_engine_inputs(rng_seed, rng_offset);` This construction only uses two parameters. According to PyTorch's code, captured_ will be set to false, and subsequent unpack in kernel_backward.h will be directly treated as a value because captured_=false. But if so, won't the pointer situation go wrong? I don't know if I have misunderstood or missed some code logic🤔

danthe3rd:
Hi,
I'm going to close this PR, as we decided to completely upstream this kernel inside PT repo (which will become the source of truth). This kernel is also available in CUTLASS, which might be a better to place to do these sort of changes:
https://github.com/NVIDIA/cutlass/blob/main/examples/41_fused_multi_head_attention/kernel_forward.h

Issue #905, opened 2023-10-24
Question: is xformers support compilation from source with cuda12.1 ?


thibautlavril:
# ❓ Questions and Help

Hello!

I am encountering some segmentation fault when trying to compile xformers from source (current main, commit "3f74d96b0d40058f036b016eb4e9e92c937cfdd9")

Python 3.8.10
torch: 2.1.0
cuda: 12.1
TORCH_CUDA_ARCH_LIST="8.6;9.0"

I tried with MAX_JOBS=2 to reduce memory pressure, but same issue

I wanted to double check that xformers was indeed supposed to support compilation from source with this CUDA version while continuing investigation on our side

Thanks!

```
...
ptxas info    : Function properties for _ZN5flash22copy_rotary_contiguousILb1ELb0EN4cute10ViewEngineINS1_8gmem_ptrIN7cutlass6half_tEEEEENS1_6LayoutINS1_5tupleIJNS9_IJNS1_1CILi8EEENSA_ILi1EEEEEENSA_ILi16EEESC_EEENS9_IJNS9_IJSC_NSA_ILi0EEEEEEjSG_EEEEES7_SJ_S7_NS8_INS9_IJNS9_IJSC_SB_EEESE_SC_EEENS9_IJNS9_IJSG_SC_EEEiSG_EEEEENS2_INS1_23ArithmeticTupleIteratorINS1_15ArithmeticTupleIJiiEEEEEEENS8_ISF_NS9_IJNS9_IJNS1_11ScaledBasisISC_Li1EEESG_EEENSU_ISE_Li0EEESG_EEEEEEEvRKNS1_6TensorIT1_T2_EERNS10_IT3_T4_EERKNS10_IT5_T6_EES1E_RKNS10_IT7_T8_EEiiii
    0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
[...]

PR #935, opened 2023-11-25
Handle CUDA 12.1 correctly in setup

arvindsun:
## What does this PR do?
Fixes Cuda 12.1 detection while building from source.

## Before submitting

- [x] Did you have fun?
  - [x] N/A
  - [x] N/A
  - [x] N/A
  - [x] N/A

## PR review

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,6 +113,8 @@ def get_cuda_version(cuda_dir) -> int:
     release = output[release_idx].split(".")
     bare_metal_major = int(release[0])
     bare_metal_minor = int(release[1][0])
+    if bare_metal_minor < 10 and len(release[1][0]) == 1:
+        bare_metal_minor = bare_metal_minor * 10
 
     assert bare_metal_minor < 100
     return bare_metal_major * 100 + bare_metal_minor


danthe3rd:
Hi,
Thanks for your PR! Can you also include the output of "nvcc -V" which caused the issue? I've been building xformers with cuda 12.1 from source for some time and it worked fine...

arvindsun:
> Hi, Thanks for your PR! Can you also include the output of "nvcc -V" which caused the issue? I've been building xformers with cuda 12.1 from source for some time and it worked fine...

nvcc output that was causing the issue: 
```
# nvcc -V
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2023 NVIDIA Corporation
Built on Mon_Apr__3_17:16:06_PDT_2023
Cuda compilation tools, release 12.1, V12.1.105
Build cuda_12.1.r12.1/compiler.32688072_0
```

As you can see `release[1][0]` had the value `1,`  in this case.

arvindsun:
> > ```
> > # nvcc -V
> > nvcc: NVIDIA (R) Cuda compiler driver
> > Copyright (c) 2005-2023 NVIDIA Corporation
> > Built on Mon_Apr__3_17:16:06_PDT_2023
> > Cuda compilation tools, release 12.1, V12.1.105
> > Build cuda_12.1.r12.1/compiler.32688072_0
> > ```
> 
> The function should return `1201` in that case. I believe this was correct before, but with your code it would return `121`, which is not what we want.
> 
> ```python
> release = "12.1,"
> print(int(''.join(filter(str.isdigit, release))))
> # 121
> ```

That is precisely the behavior I was changing! 1201 == cuda version 12.01 and 121 == cuda version 12.1 - otherwise we get GPU mismatches.

arvindsun:
> > 1201 == cuda version 12.01 and 121 == cuda version 12.1
> 
> I don't understand, this should be the same version..? What errors were you hitting? This version computed in `setup.py` is only used in a few locations:
> 
> https://github.com/facebookresearch/xformers/blob/fbf349a23a9e909863c191430bd15d5199dda14b/setup.py#L123-L132
> 
> 
> https://github.com/facebookresearch/xformers/blob/fbf349a23a9e909863c191430bd15d5199dda14b/setup.py#L150
> 
> 
> (where we expect the format to be 1201 and not 121)
> It's also used to provide a better error message in case we can't load the library:
> https://github.com/facebookresearch/xformers/blob/fbf349a23a9e909863c191430bd15d5199dda14b/xformers/_cpp_lib.py#L65-L67
> 
> 
> but none of this should cause any error

If we compile on a machine with cuda 12.1 installed (this is different than cuda 12.01), we see a version mismatch and crashes downstream for extensions not being loaded. IMO the right return value for 12.1 is 121 which will then avoid the errors discussed in https://github.com/facebookresearch/xformers/issues/905

danthe3rd:
> we see a version mismatch and crashes downstream for extensions not being loaded

The cuda version we store at build time is not used for anything that could cause this error ... If you can reproduce this error, we can have a look (you should set `XFORMERS_MORE_DETAILS=1` to get the actual error message).

The change as you propose it currently will disable the build of Flash-Attention on cuda 12.1, because it is only enabled if `cuda_version >= 1100` (https://github.com/facebookresearch/xformers/blob/fbf349a23a9e909863c191430bd15d5199dda14b/setup.py#L134)

arvindsun:
> > > we see a version mismatch and crashes downstream for extensions not being loaded
> > 
> > 
> > The cuda version we store at build time is not used for anything that could cause this error ... If you can reproduce this error, we can have a look (you should set `XFORMERS_MORE_DETAILS=1` to get the actual error message).
> > The change as you propose it currently will disable the build of Flash-Attention on cuda 12.1, because it is only enabled if `cuda_version >= 1100` (
> > https://github.com/facebookresearch/xformers/blob/fbf349a23a9e909863c191430bd15d5199dda14b/setup.py#L134
> > 
> > ).
> > This might make the error disappear, but this is also removing a big part of xFormers at the same time.
> 
> I see that now. Testing with having cuda_version return 1210 - see latest iteration of my PR - still don't know what the root cause of the build error is - will keep you posted.

BTW, do we have a wheel already built with cuda 12.1 and torch 2.1.0?

danthe3rd:
> BTW, do we have a wheel already built with cuda 12.1 and torch 2.1.0?

Yes, you can install it with:
```
pip3 install -U xformers --index-url https://download.pytorch.org/whl/cu121
```

> still don't know what the root cause of the build error is - will keep you posted.

The segmentation fault during build? Yeah we're had it occasionally. For some reason it does not happen in the CI.
But usually updating to CUDA 12.2 (which is compatible with CUDA 12.1) fixes the issue. Actually you only need to update `nvcc`.

arvindsun:
> > BTW, do we have a wheel already built with cuda 12.1 and torch 2.1.0?
> 
> Yes, you can install it with:
> 
> ```
> pip3 install -U xformers --index-url https://download.pytorch.org/whl/cu121
> ```
> 

Tried that but don't see a cu121 build here: https://download.pytorch.org/whl/xformers/

> > still don't know what the root cause of the build error is - will keep you posted.
> 
> The segmentation fault during build? Yeah we're had it occasionally. For some reason it does not happen in the CI. But usually updating to CUDA 12.2 (which is compatible with CUDA 12.1) fixes the issue. Actually you only need to update `nvcc`.

danthe3rd:
@arvindsun the wheels are in https://download.pytorch.org/whl/cu121/xformers/

arvindsun:
> https://download.pytorch.org/whl/cu121

But if you go here: https://download.pytorch.org/whl/cu121 and click on xformers it links to https://download.pytorch.org/whl/xformers

danthe3rd:
Oh yes this is a bug because the links have relative paths, and apparently
https://download.pytorch.org/whl/cu121
is not the same as
https://download.pytorch.org/whl/cu121/ (<-- Links work now)

soulteary:
It has nothing to do with the version number and does not need to be changed. @danthe3rd 

https://github.com/facebookresearch/xformers/issues/960

PR #985, opened 2024-02-23
Make Triton import conditional

sgrigory:
Windows builds are failing on Triton import in `triton_splitk.py` after it was made unconditional recently (discovered in https://github.com/facebookresearch/xformers/pull/978). This PR fixes them by making it conditional again


danthe3rd (approved):
Thanks! A bit annoying to have to pad everything tho :/
Let's make sure we catch those errors earlier in the future

Issue #1016, opened 2024-04-03
[feat] add python3.12 support


Mon-ius:
python 3.12 has unlock more power of python, and now stable with latest version `3.12.2`. 

Besides, mainstream repo including `pytorch torchvision huggingface_hub transformers accelerate diffusers has already support it`. The `xformers` is supposed to support it as well. 🤗

Mon-ius:
@danthe3rd great! if both conda and pip issue will be solved in that time?

danthe3rd:
Yes we plan to add 3.12 for both :)

bionicles:
please keep us posted if it winds up working ahead of schedule, I am excited about the py312 type system improvements for projects where i dont need to worry about backwards compatibility. The "type" keyword is crucial to distinguish variables and types and also enables us to essentially define new classes with one-liners (!) https://docs.python.org/3.12/whatsnew/3.12.html

Also, we're about to see py313 with potentially significant python performance enhancements, so it could be a good idea to think about how we might need to adjust code for python 3.13 as well in order to unlock those benefits sooner after that version is released https://docs.python.org/3.13/whatsnew/3.13.html

danthe3rd:
The PT 2.4.0 is actually a requirement to have smaller build sizes, which we need to be able to host more versions (eg py312 etc...), so I don't expect us to be ahead of schedule there.
For python 3.13, we will have to wait for PyTorch to support it first.

salcc:
Any updates on this? PyTorch 2.4 has already been released for over a month

danthe3rd:
We now have wheels for py312. Closing

Mon-ius:
@danthe3rd cheers, how about `conda` 🤗

Mon-ius:
@danthe3rd do we support python3.12 with `conda` now? 🤗

lw:
I've just changed our build scripts to target Python 3.12, but this will only have an effect next time we make a release. I don't think this change on its own is enough to warrant making a new release. Sorry.

manueldeprada:
@Mon-ius @danthe3rd any plans for 3.13 support?

PR #1017, opened 2024-04-03
add build for Python 3.12

Mon-ius:
updated the CI for conda and wheel build for Python 3.12, as requested in https://github.com/facebookresearch/xformers/issues/1016#issue-2223351721. 🤗

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -32,8 +32,8 @@ jobs:
         python:
           - "3.9"
           - "3.10"
-          - "3.11"
-          - "3.12"
+          # - "3.11"
+          # - "3.12"
         config:
           - torch_version: "2.2.2"
             torch_channel: "pytorch"
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -32,6 +32,8 @@ jobs:
         python:
           - "3.9"
           - "3.10"
+          - "3.11"
+          - "3.12"
         config:
           - torch_version: "2.2.2"
             torch_channel: "pytorch"
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -26,6 +26,7 @@ jobs:
           - "3.9"
           - "3.10"
           - "3.11"
+          - "3.12"
         torch_version:
           - "2.2.2"
         cuda_short_version:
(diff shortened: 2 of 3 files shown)


danthe3rd:
Thanks! Unfortunately the main issue here is that we are running out of storage on anaconda, and won't have space to store extra versions :/
Hopefully we can manage to significatively reduce binary size soon and support all recent python versions

Mon-ius:
Never thought it would be an issue, if that's possible to publish to pypi only?

danthe3rd:
How do you install xFormers? Do you install it with the `--index-url https://download.pytorch.org/whl/cuXXX` part in the pip install command?
If so, we don't have any space problem. We might want to store all Python versions there, and only a limited subset in pypi

Mon-ius:
we have two env setup solutions for xformer installation from `conda` and `pip`, but the `conda` one started not work due to the  version issue, that's why I am here for subimit the PR 🤗

danthe3rd:
conda is usually the better solution, but they also have the lowest storage (~10GB). We were able to [negociate](https://github.com/pypi/support/issues/2907) 20GB for pypi - trying to push it further now.
The only real solution we have is to host on the Meta servers for PyTorch for pip - which are used when you specify `--index-url https://download.pytorch.org/whl/cu121` in pip

Mon-ius:
If it is possible to increase the storage on Conda, that would be ideal. Otherwise, there will always be a compromising solution.

danthe3rd:
I'm working on getting py3.12 support for pip wheels (at least on the PyTorch s3) - hopefully we can enable that soon :)

Remember2015:
any updates?
when I use `pip install xformers --index-url=https://download.pytorch.org/whl/cu121`
it seems not work:
```
Looking in indexes: https://download.pytorch.org/whl/cu121
ERROR: Could not find a version that satisfies the requirement xformers (from versions: none)
ERROR: No matching distribution found for xformers
```

- python=3.12.3
- pytorch=2.3.0 (py3.12_cuda12.1_cudnn8.9.2_0)

danthe3rd:
I didn't get a reply, opened a new issue
https://github.com/pypi/support/issues/4168

danthe3rd:
We have builds for cp12 now, closing

PR #1041, opened 2024-05-07
The original code directly disabled the support for fused swiglu, but pytorch2.0.0 has fixed the issue in autocast of bf16.

Gstdioh:
A small PR. This PR addresses the issue of disabling fused swiglu in swiglu_op.py. The original code directly disabled the support for fused swiglu, with the comment "Let's disable autocast in bf16 until this issue is fixed". After my testing, PyTorch 2.0.0 has already fixed this bug. Therefore, we should add a judgment for the PyTorch version, and then determine whether to support fused swiglu, instead of directly disabling swiglu as before. This makes it more user-friendly.

### DIFF
--- a/xformers/ops/swiglu_op.py
+++ b/xformers/ops/swiglu_op.py
@@ -188,7 +188,7 @@ class _ForwardToPythonAutogradFunc(SwiGLUOp):
     def supports(self, op: "SwiGLUOpDispatch") -> bool:
         # Let's disable autocast in bf16 until this issue is fixed
         # https://github.com/pytorch/pytorch/issues/87979
-        if op.dtype_autocast_gpu == torch.bfloat16:
+        if op.dtype_autocast_gpu == torch.bfloat16 and torch.__version__ < "2.0.0":
             return False
         return super().supports(op)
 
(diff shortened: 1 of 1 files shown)


xformers/ops/swiglu_op.py
```diff
@@ -188,7 +188,7 @@ class _ForwardToPythonAutogradFunc(SwiGLUOp):
     def supports(self, op: "SwiGLUOpDispatch") -> bool:
         # Let's disable autocast in bf16 until this issue is fixed
         # https://github.com/pytorch/pytorch/issues/87979
-        if op.dtype_autocast_gpu == torch.bfloat16:
+        if op.dtype_autocast_gpu == torch.bfloat16 and torch.__version__ < "2.0.0":
             return False
```
danthe3rd, line 192: I believe we can remove this check all together, because xFormers now requires pytorch 2.2+
https://github.com/facebookresearch/xformers/blob/67f38b9dffc7696a6731fe3f3964369ec8113bb4/requirements.txt#L3

danthe3rd (approved):
Thanks! Can you address the comment before landing?

xformers/ops/swiglu_op.py
Gstdioh: Yeah, you're right. It can be removed. I just simply removed this check haha

danthe3rd (approved):
Thanks!

Issue #1055, opened 2024-06-07
Documentation for triton doesn't match the module


henry-paschke:
In xformers.triton, the name of the fused linear layer is FusedLinear. In this [documentation page](https://facebookresearch.github.io/xformers/tutorials/triton.html) the code block uses FusedLinearLayer, which doesn't exist.

PR #1056, opened 2024-06-07
Updated FusedLinearLayer to FusedLinear in the triton tutorial page

henry-paschke:
## What does this PR do?
Fixes issue #1055

The current documentation on triton fused linear layers used the wrong class name, FusedLinearLayer instead of FusedLinear, which is the name currently used in the module. This PR updates the docs to reflect the current name.


danthe3rd (approved):
Thanks for your contribution!

PR #1081, opened 2024-08-16
Build rocm wheels

tenpercent:
This PR enables building ROCm xformers wheels, so they can be uploaded along the CUDA xformers wheels

# Testing

Observe actions result


setup.py
```diff
@@ -417,10 +418,11 @@ def get_extensions():
             "--ptxas-options=-O2",
             "--ptxas-options=-allow-expensive-optimizations=true",
         ]
-    elif torch.cuda.is_available() and torch.version.hip:
+    elif (torch.cuda.is_available() and torch.version.hip) or (
+        os.getenv("FORCE_ROCM", "0") == "1"
```
danthe3rd, line 422: Can we simply check if `HIP_ARCHITECTURES` is non-empty here? This would avoid an additional env variable
Also, does it make sense to build for ROCM if torch does not have hip?

.github/workflows/rocm_wheels.yml
```diff
@@ -0,0 +1,66 @@
+name: rocm-wheels
+
+on:
+  push:
+    branches:
+      - develop
+  workflow_dispatch:
```
danthe3rd, line 7: Can we trigger this on PRs that might change files that are built? Similar to here: https://github.com/facebookresearch/xformers/blob/46bc6b5c5b91b4a2da3de14d0aee4f24dfe238ec/.github/workflows/win-build.yml#L4-L11
Also I would rename this file (& workload) as `rocm_build` - even though it builds a wheel, it's distinct from the wheels build+upload CI system and could be confusing otherwise.

.github/workflows/rocm_wheels.yml
```diff
@@ -0,0 +1,66 @@
...
+                    python=python,
+                    torch_version=torch_version,
+                    toolkit_type=toolkit_type,
+                    toolkit_short_version=toolkit_short_version,
+                  ))
+                  print(include[-1])
+        matrix = {'include': include}
+        print(json.dumps(matrix))
+        with open(environ["GITHUB_OUTPUT"], "a") as fd:
+          fd.write("matrix="+json.dumps(matrix))
```
danthe3rd, line 41: I don't think we need a target determinator to create the matrix, because it's static.
So I think you can put the matrix [directly](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#using-a-matrix-strategy)

danthe3rd (commented):
A few comments but this looks good!

setup.py
tenpercent: Well, we're trying to build on a CPU node, so I think `torch.cuda.is_available()` check fails. The wheels can hopefully used then on a machine with MI series GPUs. Good point about single env variable, we can do that

.github/workflows/wheels.yml
```diff
@@ -92,3 +97,23 @@ jobs:
       filter: "*torch2.4.0+cu121*"
       execute: ${{ github.repository == 'facebookresearch/xformers' && github.ref_type == 'tag' }}
 
+  upload_pt_rocm6_0:
+    needs: build
+    uses: ./.github/workflows/wheels_upload_s3.yml
+    with:
+      aws_role: "arn:aws:iam::749337293305:role/pytorch_bot_uploader_role"
+      s3_path: s3://pytorch/whl/rocm6.0/
```
tenpercent, line 105: One thing I'm not sure about is if I got the bucket path correctly. Here I'm just using the path fragment from download.pytorch.org

danthe3rd: This should be correct

danthe3rd (approved):
Thanks a lot!

danthe3rd:
So the CI started running. It seems to build, but the workers time-out after 4h I think.
Here we see that after more than 2h, it has compiled only half of it...
image

tenpercent:
> So the CI started running. It seems to build, but the workers time-out after 4h I think. Here we see that after more than 2h, it has compiled less than half of it... image

It looks like we need 5-6 hours when there are 7 compile workers. E.g. https://github.com/facebookresearch/xformers/actions/runs/10503375692/job/29096625125 finished in 5h50m. Given there is a 6 hours timeout limit, we should probably increase it. I also see that the wheels skip uploading to s3

PR #1082, opened 2024-08-17
Improvement in ROCM fmha-backward 

qianfengz:
This PR is mostly for providing update with regards to ROCM FMHA Backward.  Specifically:

1. Improved the performance of FMHA Backward for generally all input shapes
2. Added support for Headdim256 (all headdim sizes bigger than 128 and less/equal to 256 currently supported)
3. Changed to use unpadded LSE layout for var-len sequence situation
4. Improved accuracy of output grad_query for both fp16 and bf16 input type by using fp32 for AtomicAdd based accumulation
5. Adapt to the kernel API changes in ck_tile fmha fwd/bwd kernel (to support the requirements from TriDao FlashAttention)
6. Add environment variable to control the number of compiled instances
7. Add environment variable to select the using of RTZ or RTN rounding method for fp32 to bf16 conversion to balance performance and accuracy 
8. Bug fixing/stability enhancement 

To test/verify,  using the following command/scripts
```bash
#> pytest tests/test_mem_eff_attention.py::test_forward   -k "not triton" -k "not flshattF" -k "not "fa2F"
#> pytest tests/test_mem_eff_attention.py::test_backward -k "not flshattB" -k "not fa2B"
#> pytest tests/test_mem_eff_attention.py::test_dropout
#> pytest tests/test_mem_eff_attention.py::test_dropout_backward_ck
```

To benchmark performance, using 
```bash
#> python xformers/benchmark/benchmark_mem_eff_attention.py --omit-forward
```


xformers/csrc/attention/hip_fmha/attention_backward_generic_ck_tiled.cpp
```diff
@@ -515,6 +544,14 @@ efficient_attention_backward_ck(
     grad_v = tmp_grad_v_view.sum(3);
   }
 
+  /*
```
jianyuh, line 547: clean up?

xformers/csrc/attention/hip_fmha/ck_tiled_fmha_batched_backward.h
```diff
@@ -42,12 +45,18 @@ struct batched_backward_causalmask_bias_dropout_dispatch {
...
       FmhaTraits>;
 
+  static constexpr bool NeedConvertGradQ = !std::is_same<
+      typename FmhaBwdTypeConfig<ScalarType>::AccDataType,
+      typename FmhaBwdTypeConfig<ScalarType>::QGradDataType>::value;
+
   static void Run(BatchedBackwardParams& param, hipStream_t stream) {
     {
-      constexpr ck_tile::index_t kBlockSize = 256;
+      constexpr ck_tile::index_t kBlockSize = 64;
```
jianyuh, line 59: Curious about the reason on decreasing block size?

xformers/csrc/attention/hip_fmha/generate_instances.py
```diff
@@ -6,11 +6,13 @@
 #
```
jianyuh, line 6: Add a manual on how to use this generate_instances.py?

xformers/csrc/attention/hip_fmha/ck_tiled_headdim_switch.h
```diff
@@ -9,6 +9,46 @@
 #include <ck_tile/core.hpp>
 #include <stdexcept>
 
+#ifndef FMHA_SUPPORT_MAX_HEADDIM_128
```
jianyuh, line 12: Thanks for adding this! Checking if we want to only have head dim = 128 support (to save compile time), not 64, 32, 256, any easy way to configure this?

tests/test_mem_eff_attention.py
```diff
@@ -1581,7 +1608,7 @@ def test_decoder(
     # kv_heads = 1: multiquery
     # kv_heads = None: neither MQA nor GQA
     # kv_heads > 1: BMGHK
-    if dtype == "bf16" and compute_capability < (8, 0):
+    if dtype == "bf16" and torch.version.cuda and compute_capability < (8, 0):
         raise pytest.skip("BF16 is only supported on SM80+")
     import triton
```
jianyuh, line 1614: For this file, wonder if we can remove more `disable_on_rocm` for more tests after this PR?

jianyuh (commented):
for the depending Composable Kernel, is it reliable to use the current CK trunk development branch?

tests/test_mem_eff_attention.py
```diff
@@ -1003,6 +998,38 @@ def test_dropout_backward_ck(dt, q_len, kv_len, batch_size, k, p):
     )
 
 
+@cuda_only
```
tenpercent, line 1001: I think this test got here as merge conflict resolution gone bad?

xformers/csrc/attention/hip_fmha/ck_tiled_headdim_switch.h
qianfengz: Use `unset MAX_JOBS`,  the compiling is very fast.   Even though it is easy to only build for dim == 128,  we don't like do this,  since we are not very confident with our building since the scripts provided for verifying under tests/ are not specifically prepared for dim128.  You know, for any change in the codes,  we always try to run the following scripts to verify that every-thing is correctly running:
```bash
#> pytest tests/test_mem_eff_attention.py::test_forward
#> pytest tests/test_mem_eff_attention.py::test_backward
#> pytest tests/test_mem_eff_attention.py::test_dropout
#> pytest tests/test_mem_eff_attention.py::test_dropout_backward_ck
```

xformers/csrc/attention/hip_fmha/ck_tiled_fmha_batched_backward.h
qianfengz: Just keep consistent with ck_tile codes 

[fmha_bwd.py #L524](https://github.com/ROCm/composable_kernel/blob/develop/example/ck_tile/01_fmha/codegen/ops/fmha_bwd.py#L524)

which is well tested for performance consideration

xformers/csrc/attention/hip_fmha/generate_instances.py
qianfengz: Just added a `GENERATE_INSTANCES.md`

xformers/csrc/attention/hip_fmha/attention_backward_generic_ck_tiled.cpp
qianfengz: Removed the commented stuff

tests/test_mem_eff_attention.py
qianfengz: Remove the whole `test_memory_efficient_attention_full_block_masked`,  it seems this one is not used by upstream codes

danthe3rd (approved):
LGTM on the xFormers side (didn't review the generated files/generator but I trust you on that)

jianyuh (approved):

PR #1164, opened 2024-11-26
Avoid using cuda when it is installed but torch was compiled against ROCM/hip

### LINKED ISSUE #1162: Build fails agains rocm 6.2/pytorch 2.5
# 🐛 Bug

Build fails at every object with:

```
clang++: error: unknown argument: '--use_fast_math'
clang++: error: unknown argument: '--extended-lambda'
clang++: error: unknown argument: '--generate-line-info'
clang++: error: unknown argument '--threads'; did you mean '-mthreads'?
clang++: error: unknown argument: '--ptxas-options=-v'
clang++: error: unknown argument: '--ptxas-options=-O2'
clang++: error: unknown argument: '--ptxas-options=-allow-expensive-optimizations=true'
```
indeed /opt/rocm/lib/llvm/bin/clang++ (and clang in general) dose not support these options looking at setup.py it is unclear to me how this could have ever compiled against llvm.

## To Reproduce

PYTORCH_ROCM_ARCH=gfx908 python setup.py bdist_wheel

or install

## Environment

Collecting environment information...
PyTorch version: 2.5.1
Is debug build: False
CUDA used to build PyTorch: N/A
ROCM used to build PyTorch: 6.2.41134-0

OS: Arch Linux (x86_64)
GCC version: (GCC) 14.2.1 20240910
Clang version: 18.1.8
CMake version: version 3.31.0
Libc version: glibc-2.40

Python version: 3.12.7 (main, Oct  1 2024, 11:15:50) [GCC 14.2.1 20240910] (64-bit runtime)
Python platform: Linux-6.10.6-arch1-1-x86_64-with-glibc2.40
Is CUDA available: True
CUDA runtime version: 12.6.77
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: AMD Radeon RX 6800 XT (gfx1030)
Nvidia driver version: Could not collect
cuDNN version: Could not collect
HIP runtime version: 6.2.41134
MIOpen runtime version: 3.2.0
Is XNNPACK available: True

IMbackK:
fixes #1162

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -438,7 +438,7 @@ def get_extensions():
     use_pt_flash = False
 
     if (
-        (torch.cuda.is_available() and ((CUDA_HOME is not None)))
+        (torch.cuda.is_available() and (CUDA_HOME is not None) and (torch.version.cuda is not None))
         or os.getenv("FORCE_CUDA", "0") == "1"
         or os.getenv("TORCH_CUDA_ARCH_LIST", "") != ""
     ):


danthe3rd:
Thank you for the fix!

PR #1181, opened 2024-12-20
Use a constant with clearly-defined type for log2e in fwd_kernel_splitK

bertmaher:
Summary:
Triton 3.2 made some changes to its interpretation of constants (https://github.com/triton-lang/triton/pull/4613) which makes Triton more consistent with pytorch/numpy, but cause some surprising issues with this kernel.  Specifically it seems like log2e is interpreted as float32 in one instance and float64 in another, which leads to reduced prediction accuracy in some cases.

To prevent this, let's make log2e a constant and define it as float32.

### DIFF
--- a/xformers/ops/fmha/_triton/splitk_kernels.py
+++ b/xformers/ops/fmha/_triton/splitk_kernels.py
@@ -341,7 +341,13 @@ def _fwd_kernel_splitK(
     # scale sm_scale by log_2(e) and use
     # 2^x instead of exp in the loop because CSE and LICM
     # don't work as expected with `exp` in the loop
-    qk_scale = sm_scale * 1.44269504
+    #
+    # We declare log2e as a constant with a precisely-specified type to guarantee that
+    # triton will use the exact same value in all instances below, rather than sometimes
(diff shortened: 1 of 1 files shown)


bottler (approved):

xformers/ops/fmha/_triton/splitk_kernels.py
```diff
@@ -341,7 +341,8 @@ def _fwd_kernel_splitK(
     # scale sm_scale by log_2(e) and use
     # 2^x instead of exp in the loop because CSE and LICM
     # don't work as expected with `exp` in the loop
-    qk_scale = sm_scale * 1.44269504
+    log2e = tl.full((), 1.44269504, tl.float32)
```
lw, line 344: Can we add a comment with an explanation/reference for why the code is this way?

lw (approved):

xformers/ops/fmha/_triton/splitk_kernels.py
bertmaher: Yes definitely, good call

PR #1199, opened 2025-01-22
initial support blackwell

johnnynunez:
10.0 blackwell b100/b200
12.0 blackwell rtx50

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -174,8 +174,8 @@ def get_flash_attention2_nvcc_archs_flags(cuda_version: int):
         return []
     # Figure out default archs to target
     DEFAULT_ARCHS_LIST = ""
-    if cuda_version >= 1207:
-        DEFAULT_ARCHS_LIST = "8.0;8.6;8.9;9.0;9.0a;100;120"
+    if cuda_version > 1207:
+        DEFAULT_ARCHS_LIST = "8.0;8.6;8.9;9.0;9.0a;100;120;120a"
     elif cuda_version >= 1108:
         DEFAULT_ARCHS_LIST = "8.0;8.6;8.9;9.0;9.0a"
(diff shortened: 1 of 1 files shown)


danthe3rd:
Hi,
Thanks for opening this PR! Unfortunately we don't have Blackwell GPUs to test this change at the moment.
Can you detail how you tested it?
Thanks!

johnnynunez:
> Hi, Thanks for opening this PR! Unfortunately we don't have Blackwell GPUs to test this change at the moment. Can you detail how you tested it? Thanks!

Because I added support for pytorch and other libraries.
I have RTX5090
IMG_2517
IMG_2518

johnnynunez:
> How did you test xFormers with your changes? Did you run the tests? Use it in another workflow..?

Well, maybe I have to put there 

> 

Support officially comes with next cuda versions. But we have press release driver 571.86 whl, I can share you more information in the following days

johnnynunez:
Cuda 12.8 is out
whitepaper: https://docs.nvidia.com/cuda/pdf/ptx_isa_8.7.pdf
image

johnnynunez:
> Hi,
> Thanks for sharing but I don't think you addressed my question: how did you test xFormers library with these changes? Especially since you are changing the flags to build flash2, and I'm wondering what required those changes (the default is that flash2 is no longer built with xFormers, but we dispatch to the implementation in PyTorch - except for windows).

Hi,
I tested these changes internally, but I'm afraid I can't share any further details about the 5090. They work for me on both Windows and Linux.
Also, feel free to close the PR if you wish—what you mentioned is correct. However, flash attention v4 is in the works for Blackwell, since it has CUDA thread clusters similar to Hopper, so there's a new version being developed specifically for Blackwell.

PR #1202, opened 2025-01-24
update builds 12.6 and rocm 6.3

johnnynunez:
- Pytorch supports cuda 12.6 and rocm 6.2.4, 6.3
Better performance rocm6.3 https://community.amd.com/t5/ai/unlocking-new-horizons-in-ai-and-hpc-with-the-release-of-amd/ba-p/726434

```bash
pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu126
pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/rocm6.2
```

Blackwell and RDNA4 support
```bash
pip3 install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu128
pip3 install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/rocm6.3
````

PR #1204, opened 2025-01-30
Build with PyTorch 2.6.0, update CUDA targets

danthe3rd:
Taking over changes from https://github.com/facebookresearch/xformers/pull/1202 to land in xFormers to support PT 2.6.0

### DIFF
--- a/.github/actions/setup-build-cuda/action.yml
+++ b/.github/actions/setup-build-cuda/action.yml
@@ -23,16 +23,19 @@ runs:
         import sys
         print(sys.version)
         cushort = "${{ inputs.toolkit_short_version }}"
-        TORCH_CUDA_DEFAULT = "121"  # pytorch 2.4.1
+        # Version uploaded to pypi (rather than PyTorch s3)
+        TORCH_CUDA_DEFAULT = "124"  # since pytorch 2.6.0
         # https://github.com/Jimver/cuda-toolkit/blob/master/src/links/linux-links.ts
         full_version, install_script = {
-          "126": ("12.6.1", "https://developer.download.nvidia.com/compute/cuda/12.6.1/local_installers/cuda_12.6.1_560.35.03_linux.run"),
+          "128": ("12.8.0", "https://developer.download.nvidia.com/compute/cuda/12.8.0/local_installers/cuda_12.8.0_570.86.10_linux.run"),
+          "126": ("12.6.3", "https://developer.download.nvidia.com/compute/cuda/12.6.3/local_installers/cuda_12.6.3_560.35.05_linux.run"),
           "124": ("12.4.1", "https://developer.download.nvidia.com/compute/cuda/12.4.1/local_installers/cuda_12.4.1_550.54.15_linux.run"),
           "121": ("12.1.0", "https://developer.download.nvidia.com/compute/cuda/12.1.0/local_installers/cuda_12.1.0_530.30.02_linux.run"),
           "118": ("11.8.0", "https://developer.download.nvidia.com/compute/cuda/11.8.0/local_installers/cuda_11.8.0_520.61.05_linux.run"),
           "6.0": ("6.0.2", "https://repo.radeon.com/amdgpu-install/6.0.2/rhel/8.9/amdgpu-install-6.0.60002-1.el8.noarch.rpm"),
           "6.1": ("6.1.2", "https://repo.radeon.com/amdgpu-install/6.1.3/rhel/8.9/amdgpu-install-6.1.60103-1.el8.noarch.rpm"),
-          "6.2": ("6.2.3", "https://repo.radeon.com/amdgpu-install/6.2.3/rhel/8.9/amdgpu-install-6.2.60203-1.el8.noarch.rpm"),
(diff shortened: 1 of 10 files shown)


.github/actions/setup-build-cuda/action.yml
```diff
@@ -23,15 +23,19 @@ runs:
...
           "124": ("12.4.1", "https://developer.download.nvidia.com/compute/cuda/12.4.1/local_installers/cuda_12.4.1_550.54.15_linux.run"),
           "121": ("12.1.0", "https://developer.download.nvidia.com/compute/cuda/12.1.0/local_installers/cuda_12.1.0_530.30.02_linux.run"),
           "118": ("11.8.0", "https://developer.download.nvidia.com/compute/cuda/11.8.0/local_installers/cuda_11.8.0_520.61.05_linux.run"),
           "6.0": ("6.0.2", "https://repo.radeon.com/amdgpu-install/6.0.2/rhel/8.9/amdgpu-install-6.0.60002-1.el8.noarch.rpm"),
           "6.1": ("6.1.2", "https://repo.radeon.com/amdgpu-install/6.1.3/rhel/8.9/amdgpu-install-6.1.60103-1.el8.noarch.rpm"),
-          "6.2": ("6.2.3", "https://repo.radeon.com/amdgpu-install/6.2.3/rhel/8.9/amdgpu-install-6.2.60203-1.el8.noarch.rpm"),
+          "6.2.4": ("6.2.4", "https://repo.radeon.com/amdgpu-install/6.2.4/rhel/8.9/amdgpu-install-6.2.60204-1.el8.noarch.rpm"),
+          "6.3": ("6.3.1", "https://repo.radeon.com/amdgpu-install/6.3.1/rhel/8.9/amdgpu-install-6.3.60301-1.el8.noarch.rpm"),
```
lw, line 38: Do we need to keep all of these? Can we remove the ones PyTorch doesn't support anymore? (Or doesn't support yet, such as 12.8)

.github/actions/setup-build-cuda/action.yml
```diff
@@ -46,14 +50,14 @@ runs:
...
-      uses: Jimver/cuda-toolkit@v0.2.16
+      uses: Jimver/cuda-toolkit@v0.2.21
       with:
         cuda: ${{ steps.cuda_info.outputs.CUDA_VERSION }}
         method: network
 
     - name: Install python
       if: runner.os == 'Windows'
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
```
lw, line 60: Can you explain why we need to update?

xformers/components/attention/core.py
```diff
@@ -103,7 +103,7 @@ def _matmul_with_mask(
             repeat_factor = att.shape[0] // mask.shape[0]
             mask = mask.repeat([repeat_factor, 1, 1])
             logger.info("Mismatched batch dimensions for mask, repeating mask.")
-        att += mask
+        att += mask  # type: ignore
```
lw, line 106: What error was being raise here? Is there a better way to fix it?

lw (approved):
Thanks!

lw:
BTW if you want there are a couple of `# type: ignore[operator]` that can now be removed from https://github.com/facebookresearch/xformers/blob/536363ecc38809cec9c6227e6ceef88255538fea/xformers/ops/unbind.py. But we can also do it later.

xformers/components/attention/core.py
danthe3rd: https://github.com/facebookresearch/xformers/actions/runs/13053246608/job/36418024456
```
xformers/checkpoint.py:163: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
xformers/checkpoint.py:165: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
xformers/components/attention/core.py:[10](https://github.com/facebookresearch/xformers/actions/runs/13053246608/job/36418024456#step:9:11)6: error: Argument 1 to "__iadd__" of
"TensorBase" has incompatible type "Union[Tensor, SparseCS]"; expected
"Union[Tensor, Union[int, float, bool], complex]"  [arg-type]
            att += mask
                   ^~~~
Found 1 error in 1 file (checked [12](https://github.com/facebookresearch/xformers/actions/runs/13053246608/job/36418024456#step:9:13)7 source files)
```

I'm just trying to do minimal effort to get the CI green here, those components are deprecated anyway

.github/actions/setup-build-cuda/action.yml
danthe3rd: This was from https://github.com/facebookresearch/xformers/pull/1202
Probably needed for Python 3.13, but as we won't support it yet, I can revert

fialhocoelho: Are you all planning to add python>=3.13 as a requirement in the short term? I’m currently investigating when major APIs intend to introduce this limitation/update =)

danthe3rd: We won't add a requirement to python>=3.13, we want to support the same python version as PyTorch does ideally (and PT still supports 3.9 I think?)

fialhocoelho: That makes sense. Thanks for the answer!

PR #1266, opened 2025-06-16
Change schema of efficient_attention_forward_ck to have optional tensor return type for better compatibility with torch.export and AOTI 

henrylhtsang:
## What does this PR do?

Full list of changes:
1. Change schema of logsumexp to optional tensor
2. Initialize logsumexp to nullopt
3. Initialize philox_seed and philox_offset to 0
4. Symint'ify the meta function with SymInt and empty_symint (important for torch.compile as well)

## Motivation of 1 and 2

Context is https://github.com/ROCm/xformers/issues/65, but let me copy over.

Hi, we want to propose this change so the CK kernel is more compatible with pytorch (more precisely, torch.export and AOTI).

Right now, depending on `compute_logsumexp`, the second return of `efficient_attention_forward_ck`, i.e. logsumexp, is either an actual tensor or an uninitialized tensor. If it is an uninitialized tensor, it would be treated as None in Python. In other words, it functions more like an optional tensor return.

This causes some problems when using with torch.export and AOTI, since they see the schema is Tensor so didn't expect to get None.

We chatted with the xformers folks @bottler. Since the kernels are from AMD, so I want to start here. 

The change would look like the example in github.com/pytorch/pytorch/pull/154286
1. inside [attention_forward_generic_ck_tiled.hip](https://github.com/ROCm/xformers/blob/89967ff16d5103772c75160b83e756447be1e547/xformers/csrc/attention/hip_fmha/attention_forward_generic_ck_tiled.cpp#L51), change `std::tuple<at::Tensor, at::Tensor, int64_t, int64_t>` to `std::tuple<at::Tensor, std::optional<at::Tensor>, int64_t, int64_t>` for both the main and the meta functions.
2. inside [attention.cpp](https://github.com/ROCm/xformers/blob/89967ff16d5103772c75160b83e756447be1e547/xformers/csrc/attention/attention.cpp#L26), change return schema from `(Tensor, Tensor, int, int)` to `(Tensor, Tensor?, int, int)`

An alternative, as @zou3519 pointed out, is to make AOTI treat `Tensor` as `Tensor?`. I tried that route, but folks like @yiming0416 prefers modifying the schema.

## Motivation of 3

Initialize them as 0 to avoid 
```
shim_common.cpp:1264] Exception in aoti_torch: Expect returned int value to match the serialized int value, but got retured int value: 140734043882480 and serialized int value: 0
```

## Motivation of 4: Symint'ify the meta function

Following https://docs.google.com/document/d/1GgvOe7C8_NVOMLOCwDaYV1mXXyHMXY7ExoewHqooxrs/edit?tab=t.0#heading=h.1pvztxcsyq0p. Previously I have done it for https://github.com/pytorch/FBGEMM/pull/3410

Also see Richard recent PR: https://github.com/pytorch/FBGEMM/pull/4322

## Before submitting

- [x] Did you have fun?
  - [ ] Discussed offline and in github.com/pytorch/pytorch/pull/154286
  - [ ] N/A
  - [ ] N/A
  - [ ] N/A

## PR review

### DIFF
--- a/xformers/csrc/attention/hip_fmha/attention_forward_generic_ck_tiled.cpp
+++ b/xformers/csrc/attention/hip_fmha/attention_forward_generic_ck_tiled.cpp
@@ -230,11 +230,11 @@ efficient_attention_forward_ck(
 
     if (p.compute_logsumexp) {
       logsumexp = at::empty({B, Hq, M}, opts.dtype(at::kFloat));
-      p.logsumexp_ptr = logsumexp.data_ptr();
+      p.logsumexp_ptr = logsumexp->data_ptr();
       p.lse_strides = {
-          static_cast<int>(logsumexp.stride(0)),
-          static_cast<int>(logsumexp.stride(1)),
-          static_cast<int>(logsumexp.stride(2))};
+          static_cast<int>(logsumexp->stride(0)),
+          static_cast<int>(logsumexp->stride(1)),
+          static_cast<int>(logsumexp->stride(2))};
     } else {
       p.logsumexp_ptr = nullptr;
       p.lse_strides = {0, 0, 0};
@@ -379,10 +379,10 @@ efficient_attention_forward_ck(
 
     if (p.compute_logsumexp) {
       logsumexp = at::empty({1, Hq, M}, opts.dtype(at::kFloat));
-      p.logsumexp_ptr = logsumexp.data_ptr();
+      p.logsumexp_ptr = logsumexp->data_ptr();
       p.lse_strides = {
-          static_cast<int>(logsumexp.stride(1)),
-          static_cast<int>(logsumexp.stride(2))};
+          static_cast<int>(logsumexp->stride(1)),
+          static_cast<int>(logsumexp->stride(2))};
     } else {
       p.logsumexp_ptr = nullptr;
       p.lse_strides = {0, 0};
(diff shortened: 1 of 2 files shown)


henrylhtsang:
cc @tenpercent

xformers/csrc/attention/hip_fmha/attention_forward_generic_ck_tiled.cpp
```diff
@@ -48,7 +48,7 @@ namespace {
   (Mode BMHK) With all the heads having the same seqlen
   (Mode 1MHK) `batch=1` with all tokens across batches concatenated
 */
-std::tuple<at::Tensor, at::Tensor, int64_t, int64_t>
+std::tuple<at::Tensor, std::optional<at::Tensor>, int64_t, int64_t>
```
tenpercent, line 51: should we also return nullopt when logsumexp is not initialized?

tenpercent (approved):
Thanks! Please make sure it compiles before merging

henrylhtsang:
> Thanks! Please make sure it compiles before merging

I probably need to change `logsumexp.data_ptr()` to `logsumexp->data_ptr()`

henrylhtsang:
tested in fbcode, LGTM

henrylhtsang:
Actually, I might want to make two more changes, so hold on a bit more:
* use empty_symint for dynamic shapes. Since we need to symint'ify the meta function
* initialize offset and seed as 0 before returning so they are not uninitialized (thanks @yiming0416)

Will update the PR after holiday.

henrylhtsang:
re-testing for AOTI and torch.compile path in fbcode...

henrylhtsang:
> re-testing for AOTI and torch.compile path in fbcode...

done, LGTM.

@tenpercent fyi change #4 is important for torch.compile when we have dynamic shapes

tenpercent (approved):
Thanks @henrylhtsang !

Issue #1281, opened 2025-06-28
WARNING[XFORMERS]: Need to compile C++ extensions to use all xFormers features.


hanswurst232:
I'm really getting desperate, it doesn't matter what I try, I can't get xformers to work. I've tried different python versions, I've tried different torch versions ... everything fails and I always get the same message.

"WARNING[XFORMERS]: Need to compile C++ extensions to use all xFormers features.
    Please install xformers properly (see https://github.com/facebookresearch/xformers#installing-xformers)
  Memory-efficient attention, SwiGLU, sparse and more won't be available.
  Set XFORMERS_MORE_DETAILS=1 for more details"

![Image](https://github.com/user-attachments/assets/81991508-b1cd-43c3-9bb5-a7d35c261b2c)

loscrossos:
i can confirm. i am getting the same error on windows with v0.0.31. on Linux the error is not present.

# how to reproduce

create venv, activate, install torch, install xformers, then in python import torch and xformers

on windows:
```bash
py -3.12 -m venv .venv
.venv\Scripts\activate
pip3 install torch==2.7.1  --index-url https://download.pytorch.org/whl/cu128    
pip3 install -U xformers --index-url https://download.pytorch.org/whl/cu128
python -c "import torch, xformers"
```

this appears:

```
WARNING[XFORMERS]: Need to compile C++ extensions to use all xFormers features.
    Please install xformers properly (see https://github.com/facebookresearch/xformers#installing-xformers)
  Memory-efficient attention, SwiGLU, sparse and more won't be available.
  Set XFORMERS_MORE_DETAILS=1 for more details
```

if i set that variable is get this:
```
WARNING[XFORMERS]: Need to compile C++ extensions to use all xFormers features.
    Please install xformers properly (see https://github.com/facebookresearch/xformers#installing-xformers)
  Memory-efficient attention, SwiGLU, sparse and more won't be available.
Traceback (most recent call last):
  File "d:\test\.env_win\Lib\site-packages\xformers\_cpp_lib.py", line 142, in <module>
    _build_metadata = _register_extensions()
                      ^^^^^^^^^^^^^^^^^^^^^^
  File "d:\test\.env_win\Lib\site-packages\xformers\_cpp_lib.py", line 127, in _register_extensions
    raise xFormersWasNotBuiltException()
xformers._cpp_lib.xFormersWasNotBuiltException: Need to compile C++ extensions to use all xFormers features.
[...]

Issue #1272, opened 2025-06-25
Flash Attention3 not built correctly


Sakura-Luna:
# 🐛 Bug

## To Reproduce

I tested the latest build on **Windows** and it showed fa is **unavailable**. I checked the wheel content and found the filenames is wrong.

![Image](https://github.com/user-attachments/assets/80bc8973-d02d-4c5e-948d-267dea9e3a24)

## Expected behavior

After manually fixing the filenames of both files, fa works fine.

Sakura-Luna:
Although xformers removed fa2, it did not modify MemoryEfficientAttentionFlashAttentionOp accordingly, so some people may encounter errors.

danthe3rd:
Hi,
Thanks for your report! Could you clarify what filename is wrong, and what you changed it to?
Unfortunately I don't have a windows machine to test things...

Sakura-Luna:
> Hi, Thanks for your report! Could you clarify what filename is wrong, and what you changed it to? Unfortunately I don't have a windows machine to test things...

The pyd files are missing filenames, they should be _C.pyd.

danthe3rd:
Indeed I see that in the whl files... That's really strange let me check

loscrossos:
the problem is in the files:

```
.env\Lib\site-packages\xformers\pyd
.env\Lib\site-packages\xformers\flash_attn_3\pyd
```

i can confirm that renaming the files fixes the issue.


The problem comes from the `py_limited_api` feature:

https://github.com/facebookresearch/xformers/blob/ff490c3e02830b2ae0da3fbbedc96fd864d8b0c9/setup.py#L623

which was introduced here:

https://github.com/facebookresearch/xformers/commit/c35d4ea9ccb2b8084c3883fb4fbab83f6389386b#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R620

Setting it to false solves the issue as it generates the correctly named file on windows.

i have never used that feature but a quick search brings more problems than solutions on windows:

https://docs.pytorch.org/docs/stable/cpp_extension.html

https://www.google.com/search?q=py_limited_api+windows


I would propose to disable it at least on windows for the moment? 

as this went unnoticed i am guessing there are no tests run at least on the windows version?

i can not test it apart from loading on python3.12... but ABI seems to be working? maybe a windows rename job if all tests ass otherwise

iamvijan:
at 623 line i have changed it to false, but didn't worked !!!

[...]

PR #1284, opened 2025-07-02
Windows compile Fix

loscrossos:
this PR fixes the windows compile issue in #1281 and #1272. </br>It does not fix the root cause as i believe it lies in distutils handling of py_limited_api. The fix consists mitigating the issue at build time. We override the get_ext_filename function and detecting if an extension has been wrongly named and correcting it. for good measure we also control the build and local installation processes to catch any outliers. Also we detect if MSVC is being used and use the correct compiler flags for it. Tested on Blackwell

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -738,10 +738,7 @@ class BuildExtensionWithExtraFiles(BuildExtension):
             fixed_name = f"{last_part}.pyd"
             print(f"[INFO]get_ext_filename: Fixing inplace broken .pyd name: pyd -> {fixed_name}")
             return os.path.join(parent_path, fixed_name) if parent_path else fixed_name
-
         return filename
-
-
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


danthe3rd:
Hi,
Thanks a lot for your contribution! Let's merge this and check if it works better now.
I don't like that it's a hack, but I don't see another solution so far (and it's very hard for me to experiment as I don't have a windows machine at hand).

On a side note, if you have an idea of the issue, it might be worth it reporting it to distutils if you have time :)
Thanks!

loscrossos:
oh trust me, i am not happy with this either... but i spent a couple of coffe mugs too much looking for a solution.

i cant am not familiar with the feature and cant find any clue as to where to try. also its hard to test as it takes a while to compile.

i also dont even really know how to open a bug dscribing what happened. i even looked into the code of cpp extensions and distutils.

PR #1254, opened 2025-05-18
Enabling Blackwell support

loscrossos:
Nvidia Blackwell cards have been out for a while bu xformers do not formally support them. This PR adds support for them by checking for CUDA12.8 and enabling capability 120. Even though higher capabilities are available in https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#gpu-feature-list, using 120 is a conservative approach that enables all 50 series cards. This PR definitely solves #1251 and possibly #1228

Issue #1251, opened 2025-04-30
Bug with xformers 0.0.30 memory_efficient_attention on 5090 with half precision, torch 2.7, cuda 12.8


cdancette:
# 🐛 Bug

I am encountering an error with xformers 0.0.30 memory_efficient_attention on RTX 5090, with half-precision tensors. I am using torch 2.7 with cuda 12.8.

Calling the function returns the following error:
```
CUDA error (/__w/xformers/xformers/third_party/flash-attention/hopper/flash_fwd_launch_template.h:175): no kernel image is available for execution on the device
```

Note that it works in full precision.

Also, I tried installing the latest version from source  (commit 8fc8ec5a4d6498ff81c0c418b89bbaf133ae3a44), and I cannot reproduce the issue.

## Command to reproduce

```
from xformers.ops import memory_efficient_attention
import torch
q, k, v = torch.randn(5, 10, 128).half().cuda(), torch.randn(5, 10, 128).half().cuda(), torch.randn(5, 10, 128).half().cuda()
memory_efficient_attention(q, k, v)
```

Returns 
```
CUDA error (/__w/xformers/xformers/third_party/flash-attention/hopper/flash_fwd_launch_template.h:175): no kernel image is available for execution on the device
```

## Environment

You can run the script with:

- PyTorch Version (e.g., 1.0): 2.7
- OS (e.g., Linux): ubuntu 24.04.2
- How you installed PyTorch (`conda`, `pip`, source): poetry/pip
- Build command you used (if compiling from source):
- Python version: 3.12
- CUDA/cuDNN version: 12.8
- GPU models and configuration: RTX 5090
- Any other relevant information:

Cupcee:
I have same issue

silvertuanzi:
xformers 0.0.30 isn't compatible with RTX50 series. The maintainers forgot to add 10.0 (datacenter B series) and 12.0 (RTX 50 series) in `TORCH_CUDA_ARCH_LIST`.

mihkhub:
I have same issue:

```
==((====))==  Unsloth 2025.4.7: Fast Qwen3 patching. Transformers: 4.51.3.
   \\   /|    NVIDIA GeForce RTX 5090. Num GPUs = 2. Max memory: 31.345 GB. Platform: Linux.
O^O/ \_/ \    Torch: 2.7.0+cu128. CUDA: 12.0. CUDA Toolkit: 12.8. Triton: 3.3.0
[...]

PR #1285, opened 2025-07-04
adding blackwell support for precompiled wheels

loscrossos:
the current github wheel build process uses TORCH_CUDA_ARCH_LIST = 7.5 8.0+PTX 9.0a </br>this should be forwards compatible but on in some cases for blackwell it isnt (see #1251).</br> Currently the compilation code already supports blackwell.  Building with TORCH_CUDA_ARCH_LIST=12.0 solves the issue on #1251 but the pre-built wheels are not being compiled with that flag. Since the build process is done with CudaTK12.8 already i added capability 100 and 120 to the workflow. I already submitted a PR for the FA2 code path (#1254).</br>this PR adds support for blackwell on the main build workflow for the pipy builds.


bghira:
@fmassa anyone can look at this?

PR #1315, opened 2025-08-17
Change flags for FA2 and FA3

johnnynunez:
## What does this PR do?
We partially build FA2 if we want avoiding build FA3.
Also added cuda 13 if, blackwell b100/200 is compatible and partially RTX50.
Added Thor and Spark due that is SBSA


jasl:
Kindly ping for this PR

PR #1351, opened 2025-10-29
[ROCm] prepare CK sources for pytorch hipify v2 APIs

jeffdaily:
Similar to https://github.com/Dao-AILab/flash-attention/pull/1944.

See https://github.com/pytorch/pytorch/pull/151845. pytorch has removed caffe2, but hipify still contained work-arounds for caffe2 vs torch compatibility.
As a result of hipify v2 changes, some torch APIs are changing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -162,12 +162,15 @@ def get_hip_version(rocm_dir) -> Optional[str]:
 
 def detect_hipify_v2():
     try:
-        from torch.utils.hipify import __version__
         from packaging.version import Version
+        from torch.utils.hipify import __version__
+
         if Version(__version__) >= Version("2.0.0"):
             return True
     except Exception as e:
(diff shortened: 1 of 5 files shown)


bottler:
There's an update of the flash-attention submodule here to the latest commit, which includes a similar change. It looks like nothing has changed in FA2/FA3 on nvidia, so it should be fine.

jeffdaily:
@bottler could you retrigger CI after lint fix?

jeffdaily:
@bottler I had to fix a conflict.  The flash-attention submodule was updated outside of this PR.  Rebased and force pushed if you would restart CI.  Thanks.

jeffdaily:
pytorch hipify v2 changes were made backward compatible so this PR is no longer strictly needed

PR #1352, opened 2025-10-31
[NVIDIA] Fix Lint

johnnynunez:
cc @danthe3rd fix error lint


johnnynunez:
@danthe3rd linters green :)

danthe3rd:
Oh actually we already have a fix landed just now.
One thing still blocking for a release is that it looks like the win build with cuda 12.9 froze when the action finishes... https://github.com/facebookresearch/xformers/actions/runs/18955318691/job/54130560936
I've tried restarting it, hopefully it runs through this time

PR #1385, opened 2026-03-24
Make test_fwbw* explicitly require CUDA

GaetanLepage:
The tests in `tests/test_fwbw_overlap.py` fail without a GPU. They should be skipped automatically by `pytest` when `torch.cuda.is_available()` is false.

### DIFF
--- a/tests/test_checkpoint.py
+++ b/tests/test_checkpoint.py
@@ -19,8 +19,8 @@ from xformers.checkpoint import (
     list_operators,
     selective_checkpoint_wrapper,
 )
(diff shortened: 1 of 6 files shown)


GaetanLepage:
> Nice!
> 
> Though, could you import and use the `cuda_only` decorator?
> 
> https://github.com/facebookresearch/xformers/blob/ce4f89d54286ab07b90d0a230561e338eb70bff2/tests/utils.py#L25

Done! I also migrated the rest of the tests that were re-defining it unnecessarily.

PR #1398, opened 2026-06-12
tests(mac): guard fatal signal handler setup

### LINKED ISSUE #1394: chore(tests): linux error handling to be behind flags as in pytorch
https://github.com/facebookresearch/xformers/blob/ca6d2aa0d43241fc8a8dcd872debc2406160160d/tests/multiprocessing_utils.py#L70

https://github.com/pytorch/pytorch/blob/0f1785d45c0a7157e86abc214e1ee0684ef8d4a6/torch/testing/_internal/common_distributed.py#L960-L961

needed for https://github.com/NixOS/nixpkgs/pull/513735

dzmitry-lahoda:
## Summary
- Guard torch fatal-signal stack trace setup behind platform and API availability checks.
- Keep the Linux behavior while avoiding unsupported signal handler setup on Windows, macOS, and torch builds without the private helper.

Fixes #1394.

## Validation
- python -m compileall -q tests/multiprocessing_utils.py
- python -m pytest tests/test_multiprocessing_utils.py (not run: pytest is not installed in this environment)
- python -c 'import torch; print(torch.__version__)' (not run: torch is not installed in this environment)


lw (approved):
Thanks!

PR #1402, opened 2026-07-15
Fix over-counted softmax@V FLOPs for non-square causal attention

winklemad:
## What

`_attention_flops` (profiler) over-counts FLOPs for a **bottom-right causal** mask when there are more keys than queries (`Nv > N`, e.g. KV-cache decode). The "non-causal part" — the leftmost `Nv-N` key columns are visible to all `N` queries — is a dense `N x (Nv-N)` rectangle, so both the `Q·Kᵀ` and `softmax·V` matmuls run over the same `N*(Nv-N)` (query, key) pairs. The `Q·Kᵀ` term correctly uses `N * max(Nv-N, 0)`, but the `softmax·V` term uses `max(Nv-N, 0)**2`:

```python
flops = 2 * N * max(Nv - N, 0) * K + 2 * max(Nv - N, 0) * max(Nv - N, 0) * Kv
#                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ should be N * max(Nv - N, 0)
```

The non-causal `else` branch confirms the intended symmetry (`2*N*Nv*K + 2*N*Nv*Kv` — both terms share the `N*Nv` pair count).

Repro:
```python
from xformers.profiler.profile_analyzer import _attention_flops
_attention_flops([1, 1, 4, 8], [1, 1, 10, 16], causal=True)   # 1920 (over-counted), should be 1536
```

## Fix

Use `N * max(Nv - N, 0)` for the `softmax·V` pair count too. Square shapes (`N == Nv`) are unaffected — the rectangle vanishes — so the existing `test_analyze_prof_*` cases (which only use square shapes) stay green. Added `test_attention_flops_causal_nonsquare` covering the `Nv > N` path; it's a plain CPU test (no CUDA), unlike the existing `@cuda_only` profiler tests.


xformers/profiler/profile_analyzer.py
```diff
@@ -29,7 +29,7 @@ def _attention_flops(queries, values, causal: bool, fmt: str = "BHMK") -> int:
     *B, Nv, Kv = values
     if causal:  # NOTE: Causal from bottom right
         # non-causal part
-        flops = 2 * N * max(Nv - N, 0) * K + 2 * max(Nv - N, 0) * max(Nv - N, 0) * Kv
+        flops = 2 * N * max(Nv - N, 0) * K + 2 * N * max(Nv - N, 0) * Kv
```
danthe3rd, line 32: Indeed the 2 components of this formula should be symmetrical (Q@K + A@V)

danthe3rd (approved):
Thanks for catching this!

PR #1403, opened 2026-08-05
Support importing xFormers on PyTorch builds without distributed support

### LINKED ISSUE #1378: xFormers does not support running on device with `torch.distributed.is_available() = False`
# 🐛 Bug

Some devices, for instance, the NVIDIA Jetson Orin (edge compute embedded device) does not have torch.distributed built in when using the Nvidia official pytorch container for iGPU arch (e.g. `nvcr.io/nvidia/pytorch:25.03-py3-igpu`).

As a result, `torch.distributed.Work` does not exist in the customized pytorch version shipped with the container.

When `import xformers`, the type annotation on this line: 

https://github.com/facebookresearch/xformers/blob/720adff2b021f6f43957718514f5be3d10e36fb1/xformers/ops/differentiable_collectives.py#L27

will lead to `AttributeError` and cause the entire program to crash.

## Command

(Mentioned above)

## To Reproduce

(Mentioned above, specifically, this behavior persists even when the user is compiling the xFormers from source)

## Expected behavior

Similar to current framework, incompatible operators should be handled gracefully as 'not available' and allow partial usage (e.g. SDPA operators should not be affected by the missing distributed feature).

An immediate fix might be using `from __future__ import annotation` and convert the type annotation of `torch.distributed.Work` to `"torch.distributed.Work"` to prevent evaluating the type annotation during import.

## Environment

```
Collecting environment information...
PyTorch version: 2.7.0a0+7c8ec84dab.nv25.03
Is debug build: False
CUDA used to build PyTorch: 12.8
ROCM used to build PyTorch: N/A

OS: Ubuntu 24.04.1 LTS (aarch64)
GCC version: (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Clang version: Could not collect
CMake version: version 3.31.6
Libc version: glibc-2.39

Python version: 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] (64-bit runtime)
Python platform: Linux-5.15.148-tegra-aarch64-with-glibc2.39
Is CUDA available: True
CUDA runtime version: Could not collect
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: GPU 0: Orin (nvgpu)
Nvidia driver version: 540.4.0
cuDNN version: Probably one of the following:
/usr/lib/aarch64-linux-gnu/libcudnn.so.9.8.0
/usr/lib/aarch64-linux-gnu/libcudnn_adv.so.9.8.0
/usr/lib/aarch64-linux-gnu/libcudnn_cnn.so.9.8.0
/usr/lib/aarch64-linux-gnu/libcudnn_engines_precompiled.so.9.8.0
/usr/lib/aarch64-linux-gnu/libcudnn_engines_runtime_compiled.so.9.8.0
/usr/lib/aarch64-linux-gnu/libcudnn_graph.so.9.8.0
/usr/lib/aarch64-linux-gnu/libcudnn_heuristic.so.9.8.0
/usr/lib/aarch64-linux-gnu/libcudnn_ops.so.9.8.0
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                       aarch64
CPU op-mode(s):                     32-bit, 64-bit
Byte Order:                         Little Endian
CPU(s):                             12
On-line CPU(s) list:                0-11
Vendor ID:                          ARM
BIOS Vendor ID:                     NVIDIA
Model name:                         Cortex-A78AE
BIOS Model name:                    Orin Not Specified CPU @ 2.2GHz
BIOS CPU family:                    257
Model:                              1
Thread(s) per core:                 1
Core(s) per cluster:                4
Socket(s):                          1
Cluster(s):                         3
Stepping:                           r0p1
CPU(s) scaling MHz:                 100%
CPU max MHz:                        2201.6001
CPU min MHz:                        115.2000
BogoMIPS:                           62.50
Flags:                              fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp uscat ilrcpc flagm paca pacg
L1d cache:                          768 KiB (12 instances)
L1i cache:                          768 KiB (12 instances)
L2 cache:                           3 MiB (12 instances)
L3 cache:                           6 MiB (3 instances)
NUMA node(s):                       1
NUMA node0 CPU(s):                  0-11
Vulnerability Gather data sampling: Not affected
Vulnerability Itlb multihit:        Not affected
Vulnerability L1tf:                 Not affected
Vulnerability Mds:                  Not affected
Vulnerability Meltdown:             Not affected
Vulnerability Mmio stale data:      Not affected
Vulnerability Retbleed:             Not affected
Vulnerability Spec rstack overflow: Not affected
Vulnerability Spec store bypass:    Mitigation; Speculative Store Bypass disabled via prctl
Vulnerability Spectre v1:           Mitigation; __user pointer sanitization
Vulnerability Spectre v2:           Mitigation; CSV2, but not BHB
Vulnerability Srbds:                Not affected
Vulnerability Tsx async abort:      Not affected

Versions of relevant libraries:
[pip3] mypy-extensions==1.0.0
[pip3] numpy==1.26.4
[pip3] nvidia-cudnn-frontend==1.10.0
[pip3] onnx==1.17.0
[pip3] onnx_graphsurgeon==0.5.8
[pip3] onnx-ir==0.1.16
[pip3] onnxconverter-common==1.16.0
[pip3] onnxruntime==1.24.1
[pip3] onnxscript==0.6.2
[pip3] onnxslim==0.1.85
[pip3] optree==0.14.1
[pip3] pytorch-triton==3.2.0+gitb2684bf3b.nvinternal
[pip3] torch==2.7.0a0+7c8ec84dab.nv25.3
[pip3] torch_tensorrt==2.7.0a0
[pip3] torchprofile==0.0.4
[pip3] torchvision==0.22.0a0
[conda] Could not collect
```

adityasingh2400:
## What does this PR do?

Fixes #1378.

Makes xFormers importable on PyTorch builds that were compiled without distributed support, so `torch.distributed.is_available()` is `False`. This is the case for the official NVIDIA PyTorch containers for Jetson iGPU devices, for example `nvcr.io/nvidia/pytorch:25.03-py3-igpu`. Today `import xformers.ops` dies with an `AttributeError` on such a build, which takes the whole program down even for users who only want memory-efficient attention.

### The crash

When PyTorch is built without distributed support, `torch/distributed/__init__.py` skips its whole `if is_available():` block and defines only two stubs. `torch.distributed.Work`, `torch.distributed.ReduceOp` and the submodules that wrap the C++ bindings do not exist.

Repro, which also runs on a normal machine because it just removes the attribute that a real no-distributed build never has:

```
$ python -c "
import torch, torch.distributed
del torch.distributed.Work
import xformers.ops
"
Traceback (most recent call last):
  File "<string>", line 4, in <module>
  File ".../xformers/ops/__init__.py", line 29, in <module>
    from .modpar_layers import ColumnParallelLinear, RowParallelLinear
  File ".../xformers/ops/modpar_layers.py", line 11, in <module>
    from .differentiable_collectives import (
  File ".../xformers/ops/differentiable_collectives.py", line 27, in <module>
    ) -> Tuple[torch.Tensor, Optional[torch.distributed.Work]]:
                                      ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'torch.distributed' has no attribute 'Work'. Did you mean: '_Work'?
```

That is the exact traceback from the issue.

### Blast radius, and why one line is not enough

Four modules under `xformers/ops/` touch `torch.distributed` at import time, and all four are pulled in by `xformers/ops/__init__.py`:

1. `differentiable_collectives.py` annotates signatures with `torch.distributed.ProcessGroup` and `Optional[torch.distributed.Work]`. On Python older than 3.14 those annotations are evaluated while the function is being defined, so the module cannot even be created. This is the reported failure.
2. `modpar_layers.py` has the same problem with `torch.distributed.ProcessGroup`. It is the module that imports `differentiable_collectives`, so it is next in line.
3. `seqpar.py` does `from torch.distributed.distributed_c10d import _resolve_process_group, GroupName` at module level. `distributed_c10d` starts with `from torch._C._distributed_c10d import ...`, a C extension that is not built at all, so this is a hard `ModuleNotFoundError` that no annotation change can defer. On top of that, `GroupName` appears in `torch.library.custom_op` signatures, whose schemas are inferred eagerly at decoration time, so the name genuinely has to exist at import time.
4. `sequence_parallel_fused_ops.py` does `from torch.distributed._symmetric_memory import get_symm_mem_workspace`, which also reaches into `torch._C._distributed_c10d`, and it annotates its `custom_op` signatures with `dist.distributed_c10d.GroupName`.

Fixing only item 1 moves the crash to item 3 rather than resolving it. I verified this by simulating a no-distributed PyTorch with just the annotation fix in place: the next failure is `seqpar.py` line 10.

### The fix

Two parts.

Lazy annotations, which is the fix suggested in the issue. `differentiable_collectives.py` and `modpar_layers.py` get `from __future__ import annotations`, so their annotations are stored as strings (PEP 563) and evaluating `torch.distributed.Work` no longer happens at import. Nothing in either module needs annotations resolved at runtime: there is no dataclass field resolution, no pydantic, no `typing.get_type_hints` call, and no `torch.library.custom_op` decorator. Behavior when distributed is available is unchanged, and `typing.get_type_hints` still resolves the same types:

```
resolved hints: {'input_': <class 'torch.Tensor'>,
                 'process_group': <class 'torch.distributed.distributed_c10d.ProcessGroup'>,
                 'return': typing.Tuple[torch.Tensor, typing.Optional[torch.distributed.distributed_c10d.Work]]}
```

Skipping the ops that cannot exist. `xformers/ops/__init__.py` now guards the model-parallel and sequence-parallel imports with `if _HAS_TORCH_DISTRIBUTED := torch.distributed.is_available():`. These ops need NCCL collectives and symmetric memory, so they are unusable on such a build no matter what, and their modules cannot be imported at all for the reasons above. This mirrors the `_HAS_MSLK` pattern already used a few lines higher in the same file for the fmha and `rope_padded` ops, including the fact that the conditionally imported names stay listed in `__all__`. It also matches what the issue asks for: unavailable operators degrade to "not available" and partial usage keeps working.

I deliberately did not touch `seqpar.py` or `sequence_parallel_fused_ops.py`. Making them importable would mean stubbing `GroupName` and `_resolve_process_group` so that their `custom_op` registrations still succeed, which is more code and more risk for ops that can never run there anyway.

### Tests

`tests/test_import_without_torch_distributed.py` spawns a child interpreter with `sys.executable`, makes `torch.distributed` look like a build without distributed support before xFormers is imported for the first time, then imports the target module and asserts the child exits cleanly. The simulation mirrors what PyTorch itself does when `is_available()` is `False`: `is_available` returns `False`, the availability-gated attributes are removed, `GroupName` and `ProcessGroup` are replaced by stubs, and `torch._C._distributed_c10d`, `torch.distributed.distributed_c10d` and `torch.distributed._symmetric_memory` are made unimportable by putting `None` in `sys.modules`. It is parametrized over `xformers`, `xformers.ops` and `xformers.ops.differentiable_collectives`, and skips cleanly if the child cannot set the simulation up.

Note that on Python 3.14 the annotation half of this bug is hidden by PEP 649, since annotations are lazy by default there. The test still catches the import failures on 3.14 because of the eager submodule imports, and it catches the original `AttributeError` on 3.9 through 3.13, which is where the reporter is (Python 3.12).

### Fail before, pass after

Environment: macOS arm64, CPU-only, Python 3.12.12, torch 2.13.0 CPU wheel. The test only imports pure-Python modules, so the compiled extensions are not needed.

Base ref restored with `git checkout upstream/main -- xformers/ops/__init__.py xformers/ops/differentiable_collectives.py xformers/ops/modpar_layers.py`:

```
########## FAIL-BEFORE (upstream/main sources, python 3.12) ##########
E           File ".../xformers/ops/__init__.py", line 29, in <module>
E           File ".../xformers/ops/modpar_layers.py", line 11, in <module>
E           File ".../xformers/ops/differentiable_collectives.py", line 27, in <module>
E             ) -> Tuple[torch.Tensor, Optional[torch.distributed.Work]]:
E         AttributeError: module 'torch.distributed' has no attribute 'Work'. Did you mean: '_Work'?
E       assert 1 == 0
=========================== short test summary info ============================
FAILED tests/test_import_without_torch_distributed.py::test_import_without_torch_distributed[xformers.ops]
FAILED tests/test_import_without_torch_distributed.py::test_import_without_torch_distributed[xformers.ops.differentiable_collectives]
2 failed, 1 passed in 7.82s
```

Fix restored with `git checkout HEAD -- <same files>`:

```
########## PASS-AFTER (python 3.12) ##########
platform darwin -- Python 3.12.12, pytest-9.1.1, pluggy-1.6.0
collected 3 items

tests/test_import_without_torch_distributed.py::test_import_without_torch_distributed[xformers] PASSED [ 33%]
tests/test_import_without_torch_distributed.py::test_import_without_torch_distributed[xformers.ops] PASSED [ 66%]
tests/test_import_without_torch_distributed.py::test_import_without_torch_distributed[xformers.ops.differentiable_collectives] PASSED [100%]

============================== 3 passed in 5.97s ===============================
```

Same three tests also pass on Python 3.14.3.

No regression when distributed is available. Every name in `xformers.ops.__all__` that resolved before still resolves, `_HAS_TORCH_DISTRIBUTED` is `True`, and `ColumnParallelLinear`, `sequence_parallel_leading_matmul` and `fused_allgather_and_linear` are all present. The existing suites that cover the touched modules still pass or skip as they did before, on CPU:

```
$ python -m pytest tests/test_seqpar.py tests/test_sequence_parallel_fused_ops.py \
    tests/test_unbind.py tests/test_attention_patterns.py \
    tests/test_import_without_torch_distributed.py -q
159 passed, 172 skipped, 1 warning in 5.46s
```

The skips are the CUDA-only and multi-GPU cases, which this machine cannot run.

`pre-commit run --files ...` passes on every changed file with the pinned hook versions from `.pre-commit-config.yaml`, ufmt 2.8.0 with black 26.3.1 and usort 1.0.8.post1, flake8 6.1.0, mypy 1.10.0.

## Before submitting

- [x] Did you have fun?
- [x] Was this discussed/approved via a Github issue? Yes, #1378.
  - [x] N/A, no documented behavior changes on builds that have distributed support.
- [x] Did you write any new necessary tests? Yes, `tests/test_import_without_torch_distributed.py`.
- [x] Did you update the [changelog](https://github.com/facebookresearch/xformers/blob/master/CHANGELOG.md)? Yes, a `Fixed` entry under 0.0.36.


lw (commented):
Thanks for the PR! I think the change is clear and self-evidently correct, and I find the tests to be quite complicated for the value they provide. Could you remove the tests?

adityasingh2400:
Removed in 38f507f. Fair call, the test needed a subprocess with a stubbed `torch.distributed` to exercise a guard that reads plainly from the source, so it was carrying more machinery than the behavior it pinned.

The PR is now the guard itself plus the changelog entry. I checked there are no remaining references to the deleted file anywhere in the tree.

PR #1404, opened 2026-08-12
[fix] Count query heads in GQA profiler FLOPs

zupengwang:
## What does this PR do?

Correct the profiler FLOP count for grouped-query and multi-query attention.

`_attention_flops` unpacked the batch and query-head dimensions from the query shape, but then overwrote those dimensions while unpacking the value shape. When K/V heads are shared, the result was multiplied by the number of K/V heads instead of query heads. For example, an 8-query-head/2-KV-head GQA operation was under-counted by 4x, which also understated the profiler's MFU/HFU metrics.

The fix retains the query batch/head dimensions and reads only the sequence length and value dimension from the value shape. The regression covers both `BHMK` and `BMHK`, causal and non-causal paths, and MQA/GQA head counts. This changes profiler accounting only; it does not alter attention kernels or runtime execution.

No linked issue. A repository search found no existing GQA/MQA profiler FLOP report or open PR.

## Validation

- `pytest -q tests/test_profiler.py::test_attention_flops_gqa`: 8 passed
- `pytest -q tests/test_profiler.py` with CUDA hidden: 9 passed, 16 skipped
- `ufmt check`, changed-file `flake8`, changed-file `mypy`, and `git diff --check`
- Two RTX 3090 (SM86) GPUs, each tested independently with FP16 and BF16 against a float32 PyTorch reference
  - expected/formula/profiler-reported FLOPs: 134,217,728
  - previous formula: 33,554,432
  - all four outputs finite

## Before submitting

- [x] Did you have fun?
  - [ ] N/A
  - [x] N/A — no public API or usage change
  - [ ] N/A
  - [x] N/A — focused profiler correctness fix

## PR review


lw:
Can you share details about the issue you're encountering? My impression is that when using xFormers correctly your query and key tensors should have the same number of heads (possibly with the key tensor having a stride of 0 so that the same data is reused for multiple heads):
https://github.com/meta-pytorch/MSLK/blob/29fdcfcf20028faee02d89f4c45e609c88833c06/mslk/attention/fmha/common.py#L354-L360

zupengwang:
Thanks — you are right that public xFormers 4-D inputs require the same logical number of query and K/V heads; valid GQA uses 5-D tensors with explicitly expanded stride-zero K/V.

I retested this on CUDA 13 / PyTorch 2.13 across six RTX 3090s. Valid xFormers 5-D GQA is lowered by MSLK to `mslk_flash::flash_fwd`, where the profiler observes 8 physical query heads and 1 or 2 physical K/V heads. The current public head does not recognize the migrated `mslk_flash` namespace, so that valid xFormers path reports zero FLOPs.

I’ve prepared a revision that maps the MSLK Flash forward/backward operators and replaces the synthetic helper test with end-to-end SDPA and valid 5-D xFormers regressions. The targeted runtime cases pass independently on all six SM86 GPUs.

bottler:
This makes sense and fixes sdpa cases. However let's make sure we do correct the ops, like you say, so that xformers' FA2 and FA3 backends work properly. Are you planning to push a new version of this PR?

@lw Even though with GQA in fmha the caller sends 5D inputs, the mslk_flash and mslk_flash3 ops get 4D inputs, so the idea here is right for them too, even though that's currently moot.

PR #1411, opened 2026-09-20
README: note that the dev-binaries install resolves from PyPI (no --index-url)

MushiSenpai:
## What

The stable install block passes an explicit `--index-url` for each CUDA version. The **Development
binaries** block directly below it does not, while saying "Same requirements as for the stable
version above", so `pip install --pre -U xformers` resolves from PyPI instead.

PyPI ships `xformers` wheels with no CUDA local tag and `requires_dist: ['torch>=2.10', 'numpy']`,
so that command can install a wheel built against a different CUDA version than the PyTorch already
present, and can reinstall torch from PyPI over a cuda-specific or nightly build. On a Blackwell box
running a cu130 nightly torch this showed up as an NCCL ABI crash at import time, not as an install
error, which made it slow to trace back to the install command.

This adds a short note and the matching nightly-index form:

```bash
pip install --pre -U xformers --index-url https://download.pytorch.org/whl/nightly/cu130
```

Docs only, and the existing plain command is left in place.

## One thing worth flagging separately

While checking that the suggested command resolves, the nightly indexes look stale: both
`whl/nightly/cu130/xformers/` and `whl/nightly/cu128/xformers/` top out at
`0.0.34.dev20251208`, while the stable channel and PyPI are at `0.0.35`. If dev-binary publishing
is meant to be ongoing, that may be worth a look — happy to open a separate issue.


MushiSenpai:
@meta-cla check

lw:
Well there's a few things at play here.

To start, there's basically no reason anymore to want to use the development version: xFormers isn't adding any new features, and we're actively deprecating/removing most of the ones we have, hence everyone should just use a stable release. We could even remove any mention of dev builds.

Second, if I understood your issue, it comes from a mismatch in CUDA versions. This is something we're aware and have flagged to the PyTorch team (https://github.com/pytorch/pytorch/issues/187320), but there's no solution. We just can't publish a single wheel to PyPI that is compatible with all versions of PyTorch since 2.10. So far we've been procrastinating on this issue and I think we've kept compatibility with older PyTorch and thus broken it with newer versions.

Third, we're in the process of making xFormers become a pure-Python package, with no CUDA or C++ pre-built extensions. This would mean that it would not be bound to a specific CUDA version anymore. Once we land this we'll issue a new release and I expect that this would solve your problem.

MushiSenpai:
Thanks, that answers it. If dev builds are on the way out then documenting them better is the wrong
fix, and the pure-Python change sounds like it removes the whole class of problem.

Happy to do either: convert this PR to just removing the "Development binaries" block, or close it
and let the pure-Python release make it moot. Say which and I'll push.

One datapoint for the removal case: the nightly indexes look stale anyway. Both
`whl/nightly/cu128/xformers/` and `whl/nightly/cu130/xformers/` top out at `0.0.34.dev20251208`,
while stable is `0.0.35`.

For the record, what I hit was a cu130 nightly torch plus `pip install --pre -U xformers` from PyPI,
which surfaced as an NCCL ABI crash at import rather than an install error. That matches
pytorch/pytorch#187320 exactly, so no objection from me if the answer is "wait for the rewrite".

